PR #27, opened 2017-06-12
create a file with my firstname

thoat:
There's a commit by "Unknown". That was when I did not know to set up my Bash with my GitHub credentials.


jhofman:
Got it!

PR #45, opened 2019-06-11
emeka.txt

esm2000:
It took a while.


jhofman:
Close!

Do the following to put the file in the right place, from the root of the repository:

    git mv emeka.txt week1/students

Then `git commit` and `git push` your changes to your fork.

esm2000:
fixed

PR #59, opened 2021-06-01
karen.txt

KarenBritt99:


jhofman:
Awesome, got it!

Btw, is the rest of your machine already set up, or do you still have tools to install, etc.?

PR #72, opened 2021-06-02
citibike.sh answers - V

vanessamj99:


jhofman:
this like you are on the right track. as explained in the group chat, we won't actually merge these solutions into the source repo, but glad to see this.

the other important note is that you should *commit your code, not the output* to the repository. (meaning, the commands themselves, but not the actual answers.)

PR #75, opened 2021-06-03
bash scritps for citibike data analysis

Sambhav101:


week1/citibike.R
```diff
@@ -1,48 +1,103 @@
...
+# YOUR SOLUTIONS BELOW
+########################################
+
+# count the number of trips (= rows in the data frame)
+count(trips)
+
+
+# find the earliest and latest birth years (see help for max and min to deal with NAs)
+max(trips$birth_year)
+min(trips$birth_year[trips$birth_year != "\\N"])
```
jaredleekatzman, line 31: Check out https://stat.ethz.ch/R-manual/R-devel/library/base/html/Extremes.html to see how you could pass an argument to min and max to ignore NA's

week1/citibike.R
```diff
@@ -1,48 +1,103 @@
...
+  summarize(num_trips = n(), avg_time = mean(tripduration/60), sd_time = sd(tripduration/60))
+
+
+# find the 10 most frequent station-to-station trips
+trips %>%
+  group_by(start_station_name, end_station_name) %>%
+  summarize(trip_count = n()) %>%
+  arrange(desc(trip_count)) %>%
+  ungroup() %>%
+  top_n(10)
```
jaredleekatzman, line 62: tip: This works, but you could also just use head(10) for a one-line in stead of ungroup + top_n

week1/citibike.R
```diff
@@ -1,48 +1,103 @@
...
+# mutate(Date = date(starttime))
+# separate(starttime, into = c("Date"), sep = " ")
+  
+
+# compute the average number of trips taken during each of the 24 hours of the day across the entire month
+# what time(s) of day tend to be peak hour(s)?
+trips %>%
+  select(starttime) %>%
+  group_by(Hour = hour(starttime)) %>%
+  summarize(avg_trip = n()/28) %>%
```
jaredleekatzman, line 100: Why 28? always good to avoid "magic numbers" - and declare it as a variable with a name and then reference that

week1/citibike.sh
```diff
@@ -5,19 +5,61 @@
...
 # count the number of unique bikes
 
+unique_bikes=$(cut -d, -f12 201402-citibike-tripdata.csv | sort | uniq -c | awk 'NR!=1' | wc -l)
+echo -e "Number of unique bikes: $unique_bikes \n"
+
+
+
 # count the number of trips per day
 
+trips_day=$(awk -F, 'NR!=1 {counts[substr($2, 2, 10)]++} END {for (k in counts) print k"\t" counts[k]}' 201402-citibike-tripdata.csv | sort )
```
jaredleekatzman, line 22: could this be done simpler by using: cut, tr, sort, and unique?

jaredleekatzman (commented):
Looks good!

PR #98, opened 2022-05-31
"added txt"

ShanaFarber:


SoyBison:
@ShanaFarber 
You seem to have inadvertently set the line endings to windows style. Please fix this and try again. You should be able to either remove your forked branch and start over if you wish, or you may fix everything on the branch and update the pull request, that would be fine. The line endings which are used here are CRLF (\r), and we need them to be LF (\n).

ShanaFarber:
Hi, all new files that I create seem to be defaulting to CRLF format. I can open it up in visual studio code and try to convert it from CRLF to LF, but is there a way that I can globally configure the system to create LF format files when I use the console. 
I tried "touch" and "vi"

ShanaFarber:
I discovered that when I add some content to the text file and save it, it comes up as lf. I guess when it was a blank file with no text or spaces, it read it as crlf.

jhofman:
Another option is to change the way git deals with line endings globally, using this: `git config --global core.autocrlf true`

More [here](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings).

PR #100, opened 2022-06-01
Merge branch 'master' of https://github.com/msr-ds3/coursework

AR5678:


AR5678:
Cool, I was able to fix the typo. Thanks and yes NR was cool to use

Thanks for checking it over so quickly

PR #103, opened 2022-06-01
Navpreet's citibike.sh

nkaur15:


week1/citibike.sh
```diff
@@ -1,23 +1,24 @@
...
+#
+
+# count the number of unique stations
+cut -d, -f4 201402-citibike-tripdata.csv | sort | uniq | wc -l
+# count the number of unique bikes
+cut -d, -f12 201402-citibike-tripdata.csv | sort | uniq | wc -l
+# count the number of trips per day
+cut -d, -f2 201402-citibike-tripdata.csv | cut -c2-11 | sort | uniq -c
+# find the day with the most rides
+cut -d, -f2 201402-citibike-tripdata.csv | cut -c2-11 | sort | uniq -c | cort -nr | head -n1 | cut -c3-7
```
SoyBison, line 13: Check for typos before uploading in the future please.

week1/citibike.sh
```diff
@@ -1,23 +1,24 @@
...
+# find the day with the fewest rides
+cut -d, -f2 201402-citibike-tripdata.csv | cut -c2-11 | sort | grep '[0-9]' | uniq -c | sort -nr | tail -n1 | cut -c3-7
```
SoyBison, line 15: The final cut here at the end causes this line to report the number of rides instead of the day. Without that last expression it's fine so we'll leave it for today.

PR #106, opened 2022-06-01
citibike solutions

ShanaFarber:


SoyBison (approved):
Looks great! I like that you added echo output so if someone were to actually run it they'd know what's happening.

PR #108, opened 2022-06-01
solutions to citibike.sh

Sushobhan55:


SoyBison (changes requested):

week1/citibike.R
```diff
@@ -1,48 +1,82 @@
...
+nrow(trips)
+
+# find the earliest and latest birth years (see help for max and min to deal with NAs)
+summarize(filter(trips, birth_year>1), earliest_birth_year=min(birth_year), latest_birth_year=max(birth_year))
+
+# use filter and grepl to find all trips that either start or end on broadway
+filter(trips, grepl('broadway', trips$start_station_name, ignore.case = TRUE) | grepl('broadway', trips$end_station_name, ignore.case = TRUE))
+
+# do the same, but find all trips that both start and end on broadway
+filter(trips, grepl('broadway', trips$start_station_name, ignore.case = TRUE) & grepl('broadway', trips$end_station_name, ignore.case = TRUE))
```
SoyBison, line 35: Love the use of `ignore.case` obviously in this situation you don't need it but with datasets in the wild you never know how messy the data you're working on can be, so it's not a bad habit.

week1/citibike.R
```diff
@@ -1,48 +1,82 @@
...
+
+########################################
+# YOUR SOLUTIONS BELOW
+########################################
+
+# count the number of trips (= rows in the data frame)
+nrow(trips)
+
+# find the earliest and latest birth years (see help for max and min to deal with NAs)
+summarize(filter(trips, birth_year>1), earliest_birth_year=min(birth_year), latest_birth_year=max(birth_year))
```
SoyBison, line 29: that's an interesting way to filter out the `\N`s, you can also invoke the `na.rm` flag in max and min and it's a little more universal and easier to read, but this works fine.

week1/citibike.R
```diff
@@ -1,48 +1,82 @@
...
+  summarize(count=n()) %>% 
+  arrange(count) %>%
+  tail(1)
+
+# compute the average number of trips taken during each of the 24 hours of the day across the entire month
+# what time(s) of day tend to be peak hour(s)?
+trips %>% 
+  group_by(hour(trips$starttime)) %>% 
+  summarize(ave_no_trips=n()/28) %>% 
+  arrange(desc(ave_no_trips))
```
SoyBison, line 82: Basically perfect!

week1/citibike.sh
```diff
@@ -4,20 +4,29 @@
...
 
 # find the id of the bike with the most rides
+    cut -d, -f12 201402-citibike-tripdata.csv | sort | uniq -c | sort -nr | head -n1
 
 # count the number of rides by gender and birth year
+    cut -d, -f15,14 201402-citibike-tripdata.csv | sort | uniq -c | grep -v "birth"
 
 # count the number of trips that start on cross streets that both contain numbers (e.g., "1 Ave & E 15 St", "E 39 St & 2 Ave", ...)
-
+    cut -d, -f5 201402-citibike-tripdata.csv | grep '[0-9].*&.*[0-9]' | wc -l
```
SoyBison, line 28: Fantastic!

week1/citibike.sh
```diff
@@ -4,20 +4,29 @@
...
 
 # compute the average trip duration
+    awk -F\" 'NR!=1 {sum+=$2} END {print sum/NR}' 201402-citibike-tripdata.csv
```
SoyBison, line 31: Great! And you use the NR != 1 trick for killing headers!

SoyBison (approved):
Looks Great! Glad you got your login issue fixed.

week1/citibike.R
Sushobhan55: Found this way to filter NA values in the book R for Data Science.

week1/citibike.sh
Sushobhan55: That was bothering me even if it did not matter.

PR #109, opened 2022-06-01
Command Line Submission

abhipokh22:


week1/citibike.sh
```diff
@@ -4,20 +4,34 @@
...
 
 # count the number of trips per day
+cut -d, -f2 201402-citibike-tripdata.csv | cut -d' ' -f1  | sort | uniq -c | head -n-1
 
 # find the day with the most rides
+cut -d, -f2 201402-citibike-tripdata.csv | cut -d' ' -f1 | uniq -c |sort -nr | head -n1
+ 
 
 # find the day with the fewest rides
+cut -d, -f2 201402-citibike-tripdata.csv | cut -d' ' -f1 | uniq -c | sort | tail -n+2 | head -n1
```
SoyBison, line 20: Cool! I actually had to look up what the + does in tail. I'll have to use that in the future.

SoyBison (approved):
Using the +-  on head and tail is pretty slick.

week1/citibike.sh
abhipokh22: Thank you! Found out a suggestion with head -n-1 so used the same idea for tail.

PR #110, opened 2022-06-02
citibike answers

ChristopherE-crypto:
Had a hard time with the last 2 questions.


week1/citibike.sh
```diff
@@ -4,20 +4,20 @@
...
+cut -d, -f2 201402-citibike-tripdata.csv | cut -c2-11 | sort -n | uniq -c
 # find the id of the bike with the most rides
-
+cut -d, -f12 201402-citibike-tripdata.csv | sort -n | uniq -c | sort
 # count the number of rides by gender and birth year
-
+cut -d, -f14-15 201402-citibike-tripdata.csv | sort | grep '[0-9]' | uniq -c | sort
 # count the number of trips that start on cross streets that both contain numbers (e.g., "1 Ave & E 15 St", "E 39 St & 2 Ave", ...)
-
+cut -d, -f5,9 201402-citibike-tripdata.csv | grep '[0-9]' | cut -d, -f1 | grep "&" | wc -l
```
SoyBison, line 21: This unfortunately doesn’t work. Since you have the greps separated, your setup will match crosstreets like “64th and Broadway”. It may be easier here to use a single regex that could match the whole pattern. Perhaps using wildcards to handle the sections of text which are less predictable.

week1/citibike.sh
```diff
@@ -4,20 +4,20 @@
...
 
 # compute the average trip duration
```
SoyBison, line 23: I can give you a hint on this one. First separate out just the data you want, then clean it up so that it can be easily understood by a number parser. Use awk for arithmetic (using awk for arithmetic is generally a good strategy in the Unix CLI environment).

SoyBison (changes requested):
Take a look at my specific comments, and ping me on teams if you need any further help understanding.

SoyBison (approved):

SoyBison (approved):

PR #113, opened 2022-06-02
warren's answers to citibike.R

Warren-source:


week1/citibike.R
```diff
@@ -23,26 +23,52 @@ trips <- mutate(trips, gender = factor(gender, levels=c(0,1,2), labels = c("Unkn
...
 # find the earliest and latest birth years (see help for max and min to deal with NAs)
-
+#reference https://www.datasciencemadesimple.com/get-maximum-value-of-a-column-in-r-2/
+#https://stackoverflow.com/questions/652136/how-can-i-remove-an-element-from-a-list
+max(trips$birth_year)
+trips$birth_year[trips$birth_year != "\\N"] %>% min()
 # use filter and grepl to find all trips that either start or end on broadway
-
+#source https://stackoverflow.com/questions/34914173/using-grepl-to-extract-a-value-from-text-across-multiple-columns
+filter(trips, grepl('.*Broadway.*', start_station_name) | grepl('.*Broadway.*', end_station_name))
```
SoyBison, line 34: This regex works but you don't actually need the .*s, grepl matches on substrings and not the full string like some other regex matchers in other languages. This guy runs more like the re.search in python rather than re.match (which iirc must match the start of the string)

week1/citibike.R
```diff
@@ -23,26 +23,52 @@ trips <- mutate(trips, gender = factor(gender, levels=c(0,1,2), labels = c("Unkn
...
 # tip: first add a column for year/month/day without time of day (use as.Date or floor_date from the lubridate package)
-
+mutate(trips, date_self = substr(trips$starttime, 1, 10)) %>%
+ group_by(date_self) %>%
+ count() %>%
+ arrange(desc(n)) %>%
+ head(1)
 # compute the average number of trips taken during each of the 24 hours of the day across the entire month
 # what time(s) of day tend to be peak hour(s)?
+mutate(trips, hour_self = substr(trips$starttime, 12, 13)) %>%
```
SoyBison, line 71: wow you rolled your own hour parser that's slick. The one in `lubridate` I've seen as more common, but this works great!

SoyBison (approved):

PR #116, opened 2022-06-02
citibikeR exercise

ChristopherE-crypto:


week1/citibike.R
```diff
@@ -23,26 +23,47 @@ trips <- mutate(trips, gender = factor(gender, levels=c(0,1,2), labels = c("Unkn
 ########################################
 
 # count the number of trips (= rows in the data frame)
-
+nrow(trips)
 # find the earliest and latest birth years (see help for max and min to deal with NAs)
-
+max(trips$birth_year)
+trips$birth_year[trips$birth_year != "\\N"] %>% 
+ min()
```
SoyBison, line 30: This is a totally valid approach, but there are cases in other datasets where it might not work since the NANs might be recorded in a more sensible way. Luckily, `max()` and `min()` have a flag that automatically gets rid of non-numeric data which might be useful in the future.

week1/citibike.R
```diff
@@ -23,26 +23,47 @@ trips <- mutate(trips, gender = factor(gender, levels=c(0,1,2), labels = c("Unkn
...
 # find the day with the most trips
 # tip: first add a column for year/month/day without time of day (use as.Date or floor_date from the lubridate package)
-
+mutate(trips, date_self = substr(trips$starttime, 1, 10)) %>%
+ group_by(date_self) %>% count() %>%
+  arrange(desc(n)) %>%
+   head(1)
 # compute the average number of trips taken during each of the 24 hours of the day across the entire month
 # what time(s) of day tend to be peak hour(s)?
+mutate(trips, hour_self = substr(trips$starttime, 12, 13)) %>%
```
SoyBison, line 66: Nice on the substring parsing. You can also use `lubridate` 's `hour` and `floor_date` functions.

SoyBison (approved):
Looks good!

PR #117, opened 2022-06-02
Navpreet citibike.R

nkaur15:


week1/citibike.R
```diff
@@ -1,48 +1,102 @@
...
+summary <- trips %>%
+  group_by(day) %>%
+  summarize(count = n()) %>%
+  arrange(desc(count))
+head(summary,1)
+
+# compute the average number of trips taken during each of the 24 hours of the day across the entire month
+# what time(s) of day tend to be peak hour(s)?
+trips$hour <- hour(trips$starttime)
+
```
SoyBison, line 97: This is great! If for some reason in the future you come across a similar situation you can also use `mutate` to handle this in one pipeline, if say you don't want to add `hour` to your original dataframe.

SoyBison (approved):

PR #124, opened 2022-06-03
day 4 exercises

cstewart27:
I still need to complete plot_trips.R. Still resolving issues with RStudio.


SoyBison:
Looks good. No worries on the problems with RStudio, esp since we were all out of office yesterday. If you find you're still having problems, I've been having a lot of luck with setting up an RStudio server in my WSL instance, then connecting to it with a browser. Obviously it's kinda a hacky workaround and it would be better to run it natively but since so many people are having problems we may as well have it as an option.

```bash
sudo apt-get install r-base
sudo apt-get install gdebi-core
wget https://download2.rstudio.org/server/bionic/amd64/rstudio-server-2021.09.2-382-amd64.deb
sudo gdebi rstudio-server-2021.09.2-382-amd64.deb
```

then

```bash
sudo rstudio-server start
```
then connect from a web browser to:
`localhost:8787`

PR #126, opened 2022-06-03
plot_trips.R

glickdaniel98:
We ran into quite a few technical difficulties so barely got any of it done.


SoyBison:
Don't worry too much about the technical difficulties.  It's happening to a lot of students so I believe the plan is to work through it  next week.

PR #130, opened 2022-06-06
Warren's plot_trips.R

Warren-source:


week1/plot_trips.R
```diff
@@ -1,62 +1,126 @@
...
+trips %>% 
+  group_by(ymd) %>% 
+  summarize(num_trips = n()) %>% 
+  ggplot(aes(x = ymd, y = num_trips)) + 
+  geom_point() + 
+  xlab('Day') + 
+  ylab('Number of Trips')
+# plot the total number of trips (on the y axis) by age (on the x axis) and gender (indicated with color)
+trips %>% 
+  mutate(age = 2022 - birth_year) %>%
```
SoyBison, line 44: Since the data was taken in 2014, ages should be calculated from 2014 not 2022.

week1/plot_trips.R
```diff
@@ -1,62 +1,126 @@
...
+  geom_point()
+# repeat this, splitting results by whether there was substantial precipitation or not
+# you'll need to decide what constitutes "substantial precipitation" and create a new T/F column to indicate this
+mutate(trips_with_weather, heavy_weather = prcp >= .5) %>% 
+  group_by(tmin, heavy_weather) %>% 
+  summarise(count = n()) %>% 
+  ggplot(aes(x=count, y=tmin, color = heavy_weather)) +
+  geom_point()
+# add a smoothed fit on top of the previous plot, using geom_smooth
+mutate(trips_with_weather, heavy_weather = prcp >= 1.5) %>%
```
SoyBison, line 93: it's considered bad practice to use different standards in different places. I'm guessing this is just a typo or an oversight or something, but generally what I do to avoid this is, whenever there's an arbitrary constant (something I set by hand) in my code, I'll make it a variable, that way if I want to change it later, I only need to change it in one place.

week1/plot_trips.R
```diff
@@ -1,62 +1,126 @@
...
+# hint: use the wday() function from the lubridate package
+mutate(trips_with_weather, hourly = hour(starttime), day_of_week = wday(ymd)) %>%
+  group_by(day_of_week, ymd, hourly) %>%
+  summarise( count = n()) %>% 
+  group_by(day_of_week, hourly) %>%
+  summarise(avg = mean(count), std = sd(count)) %>%
+  ggplot() +
+  geom_line(aes(x = hourly, y = avg, color = "avg")) +
+  geom_line(aes(x = hourly, y = std, color = "std")) +
+  facet_wrap( ~day_of_week)
```
SoyBison, line 126: This one we talked about during the session but to reiterate:
Facet wrap is great for if you want to compare the *shapes* of data dependent on some category. It can make some things really striking too.

SoyBison (approved):
Haha! I did have stuff to say this time.

PR #131, opened 2022-06-06
plot exercise completed

Sushobhan55:


week1/plot_trips.R
```diff
@@ -1,62 +1,205 @@
...
+  group_by(ymd, tmin, substantial_prcp) %>%
+  summarize(count = n(), .groups = 'drop') %>%
+  ggplot(aes(x = tmin, y = count, label = 1:28)) +
+  labs(x = 'Minimn))um temperature',
+       y = 'No. of trips',
+       title = 'Trips tend to be low when there is substantial precipitation.') +
+  geom_point() +
+  geom_smooth() +
+  geom_text(hjust = -0.3) +
+  facet_wrap(~substantial_prcp, ncol = 1)
```
SoyBison, line 165: Our opinions differ about substantial precip haha.

week1/plot_trips.R
```diff
@@ -1,62 +1,205 @@
...
+  geom_text(hjust = -0.3) 
+std_plot_2 <- trips_with_weather %>%
+  mutate(day = wday(starttime)) %>%
+  group_by(day) %>%
+  count(ymd) %>%
+  summarize(average = mean(n), std = sd(n)) %>%
+  ggplot(aes(day, std, label = c('mon','tue','wed','thu','fri','sat','sun'))) +
+  geom_point(color = 'red') +
+  geom_text(hjust = -0.3)
+grid.arrange(ave_plot_2, std_plot_2, ncol = 1)
```
SoyBison, line 204: We weren't looking for daily averages we wanted the over the course of the day changes split up by day using facet_wrap.

SoyBison (approved):
Okay, so I think there's one main problem that all your plots are inheriting. In all of these plots, you've displayed the data in a non-standard way, which leads them to be comparatively harder to read. 
For example, instead of making two scatter plots, with points, and every point labeled, you could have just made two lines. Since the data is temporal, and therefore we know there's only y-point per x-point, producing a scatterplot only makes the plot harder to read. And not only that, the easier to read plots often can be expressed more easily in code. 
All of these criticisms fall under the "Craft" of data visualization, so they aren't the kind of thing people can just intuitively get, they take practice, and often a mentor pointing out where you could've done better.

week1/plot_trips.R
```diff
@@ -1,62 +1,205 @@
...
+# plot the minimum temperature and maximum temperature (on the y axis, with different colors) over each day (on the x axis)
+# hint: try using the pivot_longer() function for this to reshape things before plotting
+# (you can skip this and come back to it tomorrow if we haven't covered reshaping data yet)
+weather %>%
+  group_by(tmin, tmax, ymd) %>%
+  pivot_longer(c('tmin', 'tmax'), names_to = 'temperature_type', values_to = 'temperature') %>%
+  ggplot(aes(x = ymd, y = temperature, color = temperature_type)) +
+  geom_point()+
+  labs(x = 'Day',
+       title = 'minimum and maiximum temperature')
```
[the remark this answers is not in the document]
Sushobhan55: what is non-breaking typo?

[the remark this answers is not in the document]
SoyBison: It’s a typo that’s in a string or comment so it doesn’t effect how the code runs.

week1/plot_trips.R
Sushobhan55: how do you define it? i am thinking if we perform a hypothesis testing.

week1/plot_trips.R
Sushobhan55: it makes sense to me now.

week1/plot_trips.R
SoyBison: It's the amount of precipitation in inches.

PR #132, opened 2022-06-06
week 2 day 1 exercises (includes plot trips exercise)

cstewart27:
Didn't feel it was necessary to submit Rmarkdown exercises due to the nature of the exercises. No work can really be checked, it was just practice for the future. 

files to view;
plot_trips.R in week1
week_2_day_1_chapter_12_exercises.R


week1/plot_trips.R
```diff
@@ -1,62 +1,154 @@
...
+trips_with_weather %>%
+  mutate(hour = hour(starttime), wday = wday(starttime)) %>%
+  group_by(ymd, hour, wday) %>%
+  summarize(num_trips = n()) %>%
+  group_by(hour, wday) %>%
+  summarize(average_trips = mean(num_trips),
+            sd_trips = sd(num_trips)) %>%
+  ggplot(mapping = aes(x = hour)) +
+  geom_point(mapping = aes(y = average_trips, color = 'avg')) +
+  geom_point(mapping = aes(y = sd_trips, color = 'sd')) +
```
SoyBison, line 148: looks pretty good, I probably would've used a line here though, since when the quantities vary a lot over the course of the day, then scatter plots can get hard to read. Plus you don't have to worry about collisions on the X-Axis because you already grouped by hour.

week1/plot_trips.R
```diff
@@ -1,62 +1,154 @@
...
+# plot the ratio of male to female trips (on the y axis) by age (on the x axis)
+# hint: use the pivot_wider() function to reshape things to make it easier to compute this ratio
+# (you can skip this and come back to it tomorrow if we haven't covered pivot_wider() yet)
+
+trips %>%
+  group_by(birth_year,gender) %>%
+  summarize(count = n()) %>%
+  spread(gender, count) %>%
+  mutate(ratio = Male/Female) %>%
+  ggplot(aes(x = 2022-birth_year, y = ratio)) +
```
SoyBison, line 62: The data is from 2014, so you should calculate age from 2014.

week1/plot_trips.R
```diff
@@ -1,62 +1,154 @@
...
+  geom_density(mapping = aes(fill = gender))
+
+# plot the total number of trips on each day in the dataset
+ggplot(trips, mapping = aes(x = ymd)) +
+  geom_bar() +
+  scale_y_continuous(label = comma) 
+
+# plot the total number of trips (on the y axis) by age (on the x axis) and gender (indicated with color)
+trips %>% 
+  mutate(age = 2014 - birth_year) %>%
```
SoyBison, line 43: Oh I see you did 2014 here.

I find it's easiest when you are putting in arbitrary numbers (like 2014, or the precipitation boundary) to create a variable and then call it in your processing lines so that the script is consistent everywhere. Totally not a big deal here since we use it twice but if you find yourself working with huge scripts with a bunch of arbitrary boundaries it can get messy quick.

SoyBison (approved):
Looks great, the only thing I'd do differently is I'd use lineplots instead of scatter plots, they tend to look better for time series data. I also like implementing the standard deviation as a ribbon, but that would've been above and beyond for this exercise I think.

PR #133, opened 2022-06-06
week1 Day4 Exercises

ChristopherE-crypto:


week1/plot_trips.R
```diff
@@ -18,25 +24,47 @@ load('trips.RData')
 ########################################
 
 # plot the distribution of trip times across all rides (compare a histogram vs. a density plot)
-
+ggplot(data = trips, mapping = aes(x = tripduration)) + geom_histogram() + scale_x_log10(label = comma) + scale_y_log10(label = comma)
+ggplot(data = trips, mapping = aes(x = tripduration)) + geom_density() + scale_x_log10(label = comma) + scale_y_log10(label = comma)
+install.packages('patchwork')
+library(patchwork)
```
SoyBison, line 30: Two quick things about code style in R:
1. Generally R users don't like it when you put `install.packages()` in a script.
1. `library()` calls should go at the top of the file because putting them lower can cause namespace collisions in larger projects.

week1/plot_trips.R
```diff
@@ -18,25 +24,47 @@ load('trips.RData')
...
+plot1 <- ggplot(data = trips, mapping = aes(x = tripduration)) + geom_histogram() + scale_x_log10(label = comma) + scale_y_log10(label = comma)
+plot2 <- ggplot(data = trips, mapping = aes(x = tripduration)) + geom_density() + scale_x_log10(label = comma) + scale_y_log10(label = comma)
+plot1 + plot2
```
SoyBison, line 33: Not really sure what you're trying to do here since adding plots together doesn't really do anything.

week1/plot_trips.R
```diff
@@ -18,25 +24,47 @@ load('trips.RData')
...
-
+ggplot(data = trips, mapping = aes(x = tripduration)) + geom_histogram(mapping = aes(color = usertype)) + scale_x_log10(label = comma) + scale_y_log10(label = comma)
+ggplot(data = trips, mapping = aes(x = tripduration)) + geom_density(mapping = aes(color = usertype)) + scale_x_log10(label = comma) + scale_y_log10(label = comma)
 # plot the total number of trips on each day in the dataset
-
+num_rides_per_day <- group_by(trips, ymd) %>% summarize(num_rides = n())
+ggplot(num_rides_per_day, aes(x = ymd, y = num_rides)) + geom_point()
 # plot the total number of trips (on the y axis) by age (on the x axis) and gender (indicated with color)
-
+age <- mutate(trips, age = 2022 - birth_year)
```
SoyBison, line 41: The data is from 2014, so you should use 2014 to calculate age.

week1/plot_trips.R
```diff
@@ -18,25 +24,47 @@ load('trips.RData')
...
+ggplot(data = age, aes(x = age)) + geom_histogram(mapping = aes(color = gender))
 # plot the ratio of male to female trips (on the y axis) by age (on the x axis)
 # hint: use the pivot_wider() function to reshape things to make it easier to compute this ratio
 # (you can skip this and come back to it tomorrow if we haven't covered pivot_wider() yet)
+trips <- mutate(trips, age = 2022 - birth_year)
+trips_by_bd_gender <- trips %>% group_by(age, gender) %>% summarize(count = n())
+trips_by_bd_gender <- pivot_wider(trips_by_bd_gender, names_from = gender, values_from = count)
+trips_by_bd_gender <- select(trips_by_bd_gender, age, Male, Female)
+trips_by_bd_gender <- mutate(trips_by_bd_gender, ratio = Male / Female)
+ggplot(data = trips_by_bd_gender, mapping = aes(x = age, y = ratio)) + geom_point()
```
SoyBison, line 51: You kinda fell into an object oriented way of doing things here. The pipe syntax would've made this a lot cleaner.

week1/plot_trips.R
```diff
@@ -47,16 +75,33 @@ trips_with_weather <- inner_join(trips, weather, by="ymd")
...
+by_day <- group_by(trips_with_weather, date) %>% summarize(count = n())
+by_t_and_d <- group_by(trips_with_weather, ymd, tmin) %>% summarize()
+new_col <- select(by_day, count)
+x <- cbind(by_t_and_d, new_col)
+ggplot(data = x, mapping = aes(x = tmin, y = ymd)) + geom_point(aes(size = count))
 # repeat this, splitting results by whether there was substantial precipitation or not
 # you'll need to decide what constitutes "substantial precipitation" and create a new T/F column to indicate this
-
+trips_with_weather_and_p <- trips_with_weather
+trips_with_weather_and_p$sub_prcp <- with(trips_with_weather, ifelse(prcp > 1.50, "T", "F"))
```
SoyBison, line 86: This would be wayyyy easier with mutate, and also you can store the data as a boolean (as F and T without quotes) by not including the `ifelse` part and just inserting the `prcp > 1.50` directly as a column like

`trips_with_weather_and_p$sub_prcp <- trips_with_weather$prcp > 1.50`

But again, it's even cleaner if you do it with mutate in a pipe.

week1/plot_trips.R
```diff
@@ -47,16 +75,33 @@ trips_with_weather <- inner_join(trips, weather, by="ymd")
...
-
+ggplot(data = hour_total) + geom_line(mapping = aes(x = hour, y = avg)) + geom_ribbon(mapping = aes(x = hour, ymin = avg - std_dev, ymax = avg + std_dev), alpha = 0.2)
 # repeat this, but now split the results by day of the week (Monday, Tuesday, ...) or weekday vs. weekend days
 # hint: use the wday() function from the lubridate package
+trips_hour <- mutate(trips_hour, day = wday(ymd))
+trips_hour <- mutate(trips_hour, day=factor(day, levels=c(1,2,3,4,5,6,7), labels=c("Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday", "Sunday")))
+trips_by_day <- trips_hour %>% group_by(ymd, day) %>% summarize(count = n())
+day_total <- group_by(trips_by_day, day) %>% summarize(avg = mean(count), std_dev = sd(count))
+ggplot(data = day_total) + geom_point(mapping = aes(x = day, y = avg))
+ggplot(data = day_total) + geom_point(mapping = aes(x = day, y = std_dev))
```
SoyBison, line 107: This probably should've been made clear since other students did this too, but we really wanted a series of plots like the previous one, with a seperate plot for each day, that shows both average and std over the course of the day.

SoyBison (approved):
I had a couple comments, the main one for you I think is that you should embrace the pipe syntax, it would make your code a lot cleaner and it would be a lot nicer on your RAM.

PR #134, opened 2022-06-07
day4 plotting exercises

limorko:
I am uploading my exercises from the book from day 4, still working on plot_trips.R because I was having some technical difficulties, will upload that as soon as I can 

Thank you


SoyBison (approved):
Looks Good Limor

PR #144, opened 2022-06-07
Week 2 Day 2 Exercises

nkaur15:


week2/week2_day2_exercise.Rmd
```diff
@@ -0,0 +1,205 @@
...
+abs(Y - 5) <= 1
+mean(abs(Y - 5) <= 1)
+```
+
+#Question 4.1.1 - What is the value of p?
+Answer - The probabilities of the random variables of Y should add up to 1, so:
+p + 2p + 3p + 4p + 5p + 6p = 21p 
+21p = 1
+p = 1/21
+
```
SoyBison, line 38: In Rmd you can put $ around your math to make it appear better when you eventually export it.

$p + 2p$ Github has it too! You may already know LaTeX, it follows the same syntax. (roughly)

SoyBison (approved):
Great use of the Rmd features. You'll find it comes in handy a lot as we start doing more complicated things.

PR #149, opened 2022-06-08
day 6 

ShanaFarber:
I did not yet finish Friday's R script. For the last question though, what do I use the wday() function on? I don't see a column that has in it day of the week. Thanks.


SoyBison (approved):
Good use of the Rmd features!

PR #158, opened 2022-06-08
week2 day3 book exercises

ChristopherE-crypto:
There is both an RMarkdown and R file for the answers.


SoyBison (approved):
You don't really need to submit both an R script and an Rmd file in the future since I can run your code in  the Rmd on its own.

PR #169, opened 2022-06-10
Day 4 assignment 

AR5678:


week2/day4_week2.Rmd
```diff
@@ -0,0 +1,93 @@
...
+```
+c)
+The graph shows us the probability of getting -0.35 is around 0.025 (2.5%). 
+This is less than alpha so reject the null hypothesis. 
+```{r}
+prop.test(3500, 10000) ## ???
+```
+the p-value < 2.2e-16. 
+We are not rejecting the null
+
```
SoyBison, line 51: prop.test() takes a vector of successes and failures, not really sure what you're doing here since you answered the question above.

SoyBison (approved):

PR #176, opened 2022-06-10
day 9

ShanaFarber:
I tested out the "knit" function for Rmarkdown so it added a bunch of files.


SoyBison (approved):
In the future if you want to leave out files in your git commit, you can only add the ones you want, and the ones you don't care about will act like normal files. You can also do `git rm` to get rid of files you don't want and `git rm --cached` to get rid of files you don't want only on the remote, keeping them locally.

PR #177, opened 2022-06-11
week2_day5_book_exercises

ChristopherE-crypto:


week2/week2_day5_book_exercises.Rmd
```diff
@@ -0,0 +1,87 @@
...
+R <- 0.67
+
+b1 <- (sd_height / sd_shoulder_girth) * R
+b0 <- 105.2963
+
+```
+b) For the slope, for every increase of 10 cm in shoulder girth, we expect an increase in height of
+6.07675 cm on average. The intercept describes the average height (105.2963 cm) when the shoulder girth is a small value.
+
+c) R-squared is 0.4489 or 45%. R-squared is the strength of a linear fit.
```
SoyBison, line 32: "Strength of a linear fit" is a little too broad of an explanation of what R^2 is.

SoyBison (approved):

PR #179, opened 2022-06-13
daniel week3Day1

glickdaniel98:
The files are separate. There's one for the labs, one for the questions, and one for the plot/table reproduction.


week3/danielday1week3.txt
```diff
@@ -0,0 +1,40 @@
...
+
+
+
+6.3
+-----
+a-	y  = -80.41 + 0.44(gestation) - 3.33(parity) - 0.01(age) + 1.15(height) + 0.05(weight) - 8.40(smoke)
+  
+b-	Gestation: 0.44, Age: -0.01
+
+c-	Once you include other factors, it has an effect on Parity.
```
SoyBison, line 30: This is really just saying it's different, you didn't give a reason *why* it's different. This doesn't have to hit on domain knowledge, it could just be something like:
"The other variables were masking the effect from parity."

SoyBison (approved):

PR #180, opened 2022-06-13
Week 3 Day 1 submission ( Regression)

abhipokh22:


week3/Week3_day1.Rmd
```diff
@@ -0,0 +1,53 @@
...
+    
+
+6.3 
+
+a. y = -80.41 + 0.44(gestation) - 3.33 (parity) - 0.01 (age) + 1.15 (height) + 0.05 (weight) - 8.40 (smoke)
+
+b. slope of gestation = 0.44
+slope of age = -0.01 
+
+c. There might be difference because the measurement is included when holding the other variables constant.
```
SoyBison, line 35: I believe what you're getting at is what social scientists call *ceteris paribus*. The idea that because other variables are varying, the thing we're focusing on might be messed up. Another way to phrase it would be as "all else remaining equal" in the positive or "controlled variables are not being properly controlled for" in the negative.

Either way I think you're getting at the correct answer here if not stating it 100%.

week3/Week3_day1.Rmd
```diff
@@ -0,0 +1,53 @@
...
+c. There might be difference because the measurement is included when holding the other variables constant. 
+
+d. residual 
+ei = yi - y^i 
+
+```{r}
+ei = 120 - (-80.41 + 124.96 - 0.27 + 71.3 +5)
+ei
+```
+e. r^2 = 1 - var(ei)/ var(yi)
```
SoyBison, line 44: Nice use of the variance function here.

SoyBison (approved):

PR #182, opened 2022-06-13
week 3 day 1 exercises

cstewart27:


week3/week_3_day_1_exercises.Rmd
```diff
@@ -0,0 +1,247 @@
...
+#Part A
+babyWeight %>% regression_equation3 <- -80.41 + 0.44(gestation) - 0.33(parity) - 0.01(age) + 1.15(height) + 0.05(weight) + -8.40(smoke)
+
+```
+Part B 
+Age has a positive correlation and age has a negative correlation. 
+ 
+Part C
+Coefficient for parity might change as a result of introducing new variables like age. For example a female giving birth at a young age might have a baby at a heavier weight when compared to and older female giving birth. 
+
```
SoyBison, line 155: This is a good explanation, though I'd like to see a broader explanation. 

As a random nitpick: it's sometimes a pain point in scientific writing to use "female" in contexts when you're not calling attention to sex, I'd use "mother" here. That's a random nitpick though not really relevant to the statistics.

SoyBison (approved):

PR #187, opened 2022-06-13
week3_day1_book_exercises

ChristopherE-crypto:


week3/week3_day1_book_exercises.Rmd
```diff
@@ -0,0 +1,163 @@
...
+a) weight = -1.93x + 120.07
+b) Since the slope is negative, the average weight of a child who is first born is greater than the average weight of a child who is not.
+-1.93 * 0 + 120.07 = 120.07
+-1.93 * 1 + 120.07 = 118.14
+c) There is not a statistically significance between average birth weight and parity.
+
+6.3
+a) weight = 0.44x1 - 3.33x2 - 0.01x3 + 1.15x4 + 0.05x5 - 8.40x6 - 80.41
+b) The more days of gestation, the likely it is for the baby to have a higher average weight. The older the mother, the more likely it is for the baby to have a lower average weight.
+c) When adding new variables the model changes. Thus the parity
```
SoyBison, line 36: We're going to need more of an explanation than this. This phenomenon will not occur in all situations, and you should consider what is special about this data that is causing this phenomenon. Appeal to how these variables are related to one another, and how they *could* be related to one another. Bonus points if you have a domain-specific explanation.

SoyBison (approved):

PR #188, opened 2022-06-14
day 10

ShanaFarber:
I could not resolve the errors for lab 5.3.1. I received an error when trying attach(Auto) that objects were hidden and I thought to detach the Carseats data or close down and restart Rstudio but it did not work. I then got subsequent errors in each of the mean((...)) portions but I wasn't sure if maybe this was stemming from the original error.


week3/day_10_exercises.R
```diff
@@ -0,0 +1,199 @@
...
+#>(b) Interpret the slopes of gestation and age in this context.
+#>      The slope for gestation here indicates a predicted increase of 0.44 ounces for every
+#>      day of gestation. The slope for age shows a predicted decrease of 0.01 ounces for every 
+#>      year of the mother's age. 
+#>      
+#>(c) The coefficient for parity is different than in the linear model in exercise 6.2. 
+#>    Why might there be a difference?
+#>      This might be bc the pattern of effect is different when comparing just bwt against parity
+#>      vs. in addition to all the other variables. I.e. The relationship between parity and bwt is different
+#>      than the relationship between bwt and parity in addition with the other variables in the dataset.
```
SoyBison, line 130: This is a pretty good explanation, but it would be nice to also get into why it's different, and what it is about the new variables that could be causing this change.

SoyBison (approved):
I think you have to attach the datasets by name, but don't worry too much about it you did well on everything else.

PR #211, opened 2022-06-16
yesterdays week3day3

glickdaniel98:


week3/ngrams/05_final_report.Rmd
```diff
@@ -169,6 +232,13 @@ Write up your answer to Part F here.
 
 Write up your answer to Part G here. Include code that shows the years with the smallest and largest half-lifes.
 
+Yes. In the 1850's and 1900's. Maybe because of big wars?
```
SoyBison, line 235: Not a bad explanation I think. We get the American Civil war in the 1860s, and the european revolutions in 1848. The late 1910s early 1920s obviously were not a fantastic time for printing books either.

SoyBison (approved):

PR #212, opened 2022-06-16
warrens week 3 day 4

Warren-source:


week3/predict_citibike.Rmd
```diff
@@ -0,0 +1,176 @@
...
+ggplot(plot_data, aes(x=K, y=error, color=split)) +
+  geom_line() +
+  scale_x_continuous(breaks=K) +
+  xlab('Polynomial Degree') +
+  ylab('RMSE')
+
+
+```
+
+Yeah these are pretty similar results, though I'm not sure why the lables are reversed. Now I'm gonna do my thing with the linear model.
```
SoyBison, line 94: The labels are correct, but their being switched in terms of rank is not a big deal. We are working with a pretty small validation set, so weird things can happen like this.

week3/predict_citibike.Rmd
```diff
@@ -0,0 +1,176 @@
...
+plot_data <- bind_rows(trips_per_day_train, trips_per_day_validate)
+
+ggplot(plot_data, aes(x = ymd, y = num_trips)) +
+  geom_point(aes(color = split)) +
+  geom_line(aes(y = pred)) +
+  xlab('Day') +
+  ylab('Number of Trips') 
+```
+
+This looks really overfit. But I think this is due to the inclusion of weekdays factor. Since travel would go up and down based on if it's a weekday or day.
```
SoyBison, line 167: It probably only looks overfit on this kind of plot because we're plotting with respect to time, and time isn't part of the model (really, the `weekday` variable is time-based, but as you observed, it just makes the plot look more sporadic.)

SoyBison (approved):

PR #214, opened 2022-06-16
Week 3 day 4 exercises

cstewart27:


week3/predict_citibike.Rmd
```diff
@@ -0,0 +1,167 @@
...
+
+plot_data <- bind_rows(trips_per_day_train, trips_per_day_validate)
+ggplot(plot_data, aes(x = ymd, y = num_trips)) +
+  geom_point(aes(color = split)) +
+  geom_line(aes(y = pred)) +
+  xlab('Day') +
+  ylab('Number of Trips') 
+```
+
+Although the it appears to represent some overfit, that could be a result of weekday factors since travel is dependent upon day of the week.
```
SoyBison, line 157: Yeah it probably just looks like that because "time" is not really part of your model, so there's other relationships, implicit and explicit that are being exaggerated.

SoyBison (approved):

PR #217, opened 2022-06-17
Predict Citibike

nkaur15:


week3/predict_citibike.RMd
```diff
@@ -0,0 +1,370 @@
...
+ggplot(plot_data, aes(x=K, y=avg_validate_err)) +
+  geom_pointrange(aes(ymin=avg_validate_err - se_validate_err,
+                      ymax=avg_validate_err + se_validate_err,
+                      color=avg_validate_err == min(avg_validate_err))) +
+  geom_line(color = "red") +
+  scale_x_continuous(breaks=1:12) +
+  theme(legend.position="none") +
+  xlab('Polynomial Degree') +
+  ylab('RMSE on validation data')
+```
```
SoyBison, line 221: Why do you think this plot turned out flat?

SoyBison (approved):
I think you just need more writing in here, it'd be nice to see what you're thinking as you go through all of these different explorations. That being said I like the section where you're trying a bunch of different models, but it would be nice to see what you're thinking.

PR #218, opened 2022-06-17
Week 3 Day 4 Classwork

ElissaLeung:


week3/predict_citibike.Rmd
```diff
@@ -0,0 +1,264 @@
...
+
+
+1. Trips per day file
+
+```{r}
+
+trips_per_day <- read_tsv('trips_per_day.tsv') %>%
+  mutate(day = wday(ymd), weekday = (day < 6))
+
+holiday <- read_csv('holidays.csv', col_names = FALSE)
```
SoyBison, line 29: If you include outside data you should include it, or link to it so that people can run your code easily. If that's not an option, you can use the knit feature of RStudio to make a copy of your exploration that people can read without running your code.

SoyBison (approved):

PR #220, opened 2022-06-17
Citibike Prediction

nirvi8:


SoyBison:
Looks good in general. I'm going to repeat the thing about including holidays.csv that we talked about in the breakout room today. Some more explanation of the plots would be nice in the future as well. Also comments that narrate why you're adding objects to the model that you are would make it a lot easier to read. Besides those little things it's good though!

PR #221, opened 2022-06-17
week2day4 exercises

glickdaniel98:
still need to use it on test set.


SoyBison (approved):
Looks good! I'd appreciate some comments on your model as you're building it, but besides that it's great!

PR #223, opened 2022-06-17
week3_day4_citibike_predict

ChristopherE-crypto:


SoyBison (approved):
I like your choice of going for a simpler model. You did a good job annotating what you're doing as well.

PR #250, opened 2023-05-31
Finished citibike.sh

jerikjakobsen:


epickens:
Looks good!

As a challenge, can you find a way to simplify the first solution?

PR #251, opened 2023-05-31
Update citibike.sh

alexkhuey:


epickens:
These look like valid solutions! I think you have the right idea. 

As far as finding a "day" goes, you can simplify your commands by noticing that day = date when working with a single month.

PR #253, opened 2023-05-31
jerry day2

jchen056:
pull request day2:  command lines


epickens:
Looks pretty good. 

Here are a few suggestions:

- instead of using grep to break on spaces you can use "cut"
- for the first question, you don't need to use awk. One of the columns contains station name. You can count the number of unique stations using that column

PR #255, opened 2023-06-01
 Solved citibike problems

SaraGoldb:
Solved citibike problems


epickens:
I think you have the right idea, but some of the solutions are a bit more complicated than they need to be. 

Can you figure out how to do the first two without awk? Hint: do you need to use multiple columns or can you get away with one?

Also, do you need  tr -d '"'  to questions 7 & 8?

SaraGoldb:
For the first and second question I did it with awk because I thought that was the fastest way, since it cuts out sort. Is this not the case?
For the first question I can also do
`cut -d, -f4 201402-citibike-tripdata.csv | sort | uniq | wc -l`
But isn't it true that if there is a unique station in the end_station colmn that does not appear in the first that this station will not be counted? Is the likeliness of this so low that we just don't care, but in any case how _would_ you account for that?
For the second I can also just do
`cut -d, -f12 201402-citib
ike-tripdata.csv | sort | uniq | wc -l`

I don't need tr -d '"' for questions 7 & 8, I just didn't like seeing the quotes around the data.

Thank you!

PR #258, opened 2023-06-01
Command line practice

Waraba2:


epickens:
Things are looking good so far. Just finish up those last few. No real comments on the others though!

PR #259, opened 2023-06-01
Completed citibike.R assignment

alexkhuey:


epickens:
Take a second look at the second to last R question. Each day we can observe the number of trips in each hour. We can then do that for each day in the month. That should give us a months worth of examples of the number of trips during each hour. We can then use that info to get the final solution.

PR #260, opened 2023-06-01
Finished R Questions

Elvis74751:


epickens:
Looks pretty good. 

Take a second look at the second to last question. What do we need to find the average number of trips per hour? We need a set of examples of trips per hour that we can average. 

Try rewriting the code to get samples for each hour that you can then average. I.E. for hour x get (100, 150, 111,...) that you can then average.

PR #263, opened 2023-06-01
Jerry week1 day3 R

jchen056:
R day1 data manipulation


epickens:
Looks good. Seems like you became a lot more efficient as you progressed through the assignment!

PR #264, opened 2023-06-01
Citibike R assignment

10zinchosang:


epickens:
Looks pretty good. 

Here are a couple notes: 

For "find the top 3 end stations for trips starting from each start station" you keep a (start, end) grouping throughout your solution. You need to switch to a (start) only grouping to get the final answer.

Also for the last question you don't need to use n()/28. You can use mean(). To make this work, remember that you can call multiple functions within a single summarize (e.g. summarize(x=...,y=...)).

PR #267, opened 2023-06-02
completed citibike.R 

XiuwenZ:


epickens:
Looks pretty good so far. Try to finish the remaining questions when you get the chance.

Also, for "find the top 3 end stations for trips starting from each start station" you need to do a second group_by to break out of the (start, end) grouping and use a (start) only grouping to get the final correct result.

PR #268, opened 2023-06-02
Command Line exercises on citibike.sh (by Albina Haque)

Albina75:
In Ubantu (command line exercise) the command for the average (last question) is not working. Iam confused about that .


epickens:
Looks good.

PR #270, opened 2023-06-02
citibike.R

hadassahkk:
bibata and hadassah


epickens:
Looks pretty good!

My only comment is that I think you misread "find the top 3 end stations for trips starting from each start station." Trying doing it again to get the top 3 per start station instead of the top three overall.

PR #272, opened 2023-06-05
Added to plot_trips.R

SaraGoldb:
I solved 'plot trip data' and 'plot weather data'. I began solving 'plot trip and weather data' but I don't really understand how to do them.


jhofman:
No worries---you'll have time to work on the rest later today!

PR #273, opened 2023-06-05
ggplot practice

Waraba2:


epickens:
Good start to the plots!

epickens:
Good work overall. 

Take a second look at the questions that ask you to plot trip count as a function of the weather. To clarify the question, saying we want trips as a function of weather means we want the weather statistic on the x-axis and the count on the y-axis. You're very close. Try to alter your answer slightly to get that different x vs y.

PR #280, opened 2023-06-06
added plot_trips.sh

XiuwenZ:


epickens:
For the questions like "plot the number of trips as a function of the minimum temperature/precipitation/etc." you're currently grouping by date and weather, getting counts and then plotting counts against the date. This means you've plotting the number of trips as a function of date instead of weather. In fact, given that each day has just one weather statistic (tmin for example), grouping by ymd and weather will give us the same counts as just grouping by ymd. What you want instead is to find a way to plot the number of trips on the y-axis and the weather statistic on the x-axis.

Also, Please give more detailed commit messages in the future so I know exactly what changes I should be looking for!

PR #282, opened 2023-06-13
wk3 d2 exercises

jchen056:
regression+rmarkdown


epickens:
Looking good so far, I'm not seeing anything I would change.

PR #283, opened 2023-06-15
Jerry Week3 Day4

jchen056:
week3 day4 long tail exercises: (last one off about by 5000 for inventory size)


epickens:
Looks great!

For readability I would suggest that you create new lines when doing multiple dplyr operations. I just makes it easier to read for both yourself and others (like me) that might be reading or using your code.

A few examples would be:

Data %>%
   operations1() %>% 
   operation2()...

Data %>% 
    mutate(x = c1,
                y = c2,
                z = ...) %>%
    summarise(s1 = u,
                       s2 = v,
                       s3 = ...)

PR #287, opened 2023-06-19
Completed week3 movielens and ngrams 01-05 files

SaraGoldb:


epickens:
Everything is looking good. 

As Jake said in class today, using filter for max(prop) and min(half_prop) can give some weird behavior that you need to watch out for, but it's by no means incorrect.

PR #305, opened 2024-06-04
solutions to citibike problem set

salma-ghazii:


jennyw23:
hi salma! there is one main question to recheck, # count the number of trips that start on cross streets that both contain numbers (e.g., "1 Ave & E 15 St", "E 39 St & 2 Ave", ...). 

I also want to ask if you can think of another way of cutting the timestamp to keep the date; instead of indexing using `cut -c1-10`, how could you get the date if there wasn't always consistently 10 date characters? (e.g. if sometimes it said 02/01/2024 and other times 02/01/24?

PR #306, opened 2024-06-04
Solved Bash Problems

Songblade:
Salma and I solved the bash problems, and would like you to look them over.


jennyw23:
good work overall! I commented this to Salma too, but check line 21 (the street cross sections)!

PR #309, opened 2024-06-04
This is my day 2 work

ubaydulsami:


jennyw23:
good work! although NR works well here, you can also save a variable that counts the number of rows you are iterating over with something like this:

`awk '{total += $1; n++;} END { print (total / n) }`

PR #310, opened 2024-06-04
edit the citibike.sh file

YinoXu:


jennyw23:
good work! two main comments: this dataset was already sorted in order by date, so it wasn't necessarily to sort first before getting unique values (if you remember what Jake said in class, the sort function doesn't sort the way that Python does. It just checks that the value before it is the same). If the dataset wasn't in date order, question 2 might not have worked! 2) good job with the cross streets regular expression! remember to use `wc -l` to get the count :)

PR #312, opened 2024-06-04
Solved Bash Problems (Redux)

Songblade:
I have updated the incorrect problems to be correct, and added a calculation of variance. If it seems high, note that some people took the bikes for a week at a time, and this is variance of minutes.


jennyw23:
good work! on the note of high values, something to think about is what might have gone wrong if the absolute value of the numbers gets very very large, but the variance stays small (like if the mean = 1e12 but sd = 1)

PR #313, opened 2024-06-04
answers for citibike.sh

phuongjanele:


jennyw23:
good job! The one thing to check is the pattern in your grep `cut -d, -f5 201402* | grep '[0-9].*[0-9]' | sort | uniq | wc -l`

if you look at the output before `wc -l`, you'll see that '[0-9].*[0-9]' sometimes matches with cross sections like W 16 St & The High Line, which have two numbers on only one street. Hint: what if you use the & sign in the middle to separate the two streets?

PR #314, opened 2024-06-04
citibike completed

fabio-pecora:


jennyw23:
good job! and right idea with the last solution, but you can also do it more concisely! `awk` is really powerful because it can be written in a single line. For instance, Jake used awk like this: `awk -F, '{counts[$15]++} END {for (k in counts) print counts[k]"\t" k }' 201402-citibike-tripdata.csv` to count by gender. You can also use awk to count up rows using this syntax `VAR+=value`. And, you can use semicolons to separate commands that would normally go line after line in python. Assuming you start with this: `awk '{total += $1; count++;} ...'`, how would you get the average?

fabio-pecora:
Thanks for the feedback! Would it be END {print total/count)’ 201402-   ?

Fabio

jennyw23:
yep! `awk '{total += $1; count++;} END {print total/count)' 201402-` (don't forget the * to complete the file name!)

PR #317, opened 2024-06-05
Update citibike.sh

suvedeisoyol:


jennyw23:
A few notes! To do counting, try using `wc`, which means "word count". This will spit out a number more easily than `cat -n`.
The output of "find the day with the most and fewest rides" is correct, but if you want to output only the value, try using `tail -n1 and head -n1` to list only the top value! (same for bike id with most rides)

Your solution to compute the average trip duration is clever, but does a lot of repeated summing! The follow-up message you sent me is a simpler/more efficient solution :) but both produce the right answer!

PR #321, opened 2024-06-05
Solved R Problems

Songblade:


jennyw23:
good work (remember to start from the data then pipe onward (`trips %/% group_by .....`) to make the code cleaner!

PR #325, opened 2024-06-05
day 3 work submission

ubaydulsami:


jennyw23:
good job! Another possible way of doing the first question in the pipe format is to do trips %>% and then a summarize. Can you guess what function you'd use to just get the total rows?

For the "start / end of broadway" questions, you used the `^` to explicitly find streets where Broadway was the first word in the station names! The `^` is super useful for specifically finding keywords that start in the beginning of a string. You can remove the `^` to get stations that contain the word "Broadway" even if it isn't the first word in the station name. 

The answer here: 
`trips %>% 
  group_by(gender) %>% summarize(count = n(),
                                 mean(tripduration), 
                                 sd(tripduration))`

is correct, but make sure to set a column name (exactly how you did with count) so that we can identify mean(tripduration) and sd(tripduration). just include `mean_tripduration=mean(tripduration)`, etc. to do this :) 

In the top 3 most common station-to-station trips, it wasn't totally necessary to add gender into the arrange function before you grouped by gender since you're grouping by gender after! Normally, I would group by gender first, then try to arrange the counts after I've grouped it, then slice the head of each group. 

trips %>% 
  group_by(gender, start_station_name, end_station_name) %>% 
  summarise(count = n()) %>% 
  arrange(gender, desc(count)) %>% 
  group_by(gender) %>% 
  slice_head(n = 3)

ubaydulsami:
**suggestion:** Another possible way of doing the first question in the pipe format is to do trips %>% and then a summarize. Can you guess what function you'd use to just get the total rows?

**reply:** Yes, I can use summarize and then count using "n ()" to get the same result. I have updated my code according to that. 

**suggestion:** For the "start / end of broadway" questions, you used the ^ to explicitly find streets where Broadway was the first word in the station names! The ^ is super useful for specifically finding keywords that start in the beginning of a string. You can remove the ^ to get stations that contain the word "Broadway" even if it isn't the first word in the station name.

**reply:** There might be confusion here, but I have not used '^' once in my code

**suggestion:** The answer here:
trips %>%  group_by(gender) %>% summarize(count = n(), mean(tripduration),  sd(tripduration))

is correct, but make sure to set a column name (exactly how you did with count) so that we can identify mean(tripduration) and sd(tripduration). just include mean_tripduration=mean(tripduration), etc. to do this :)

**reply:** I actually did do the second way, my code is below:

trips %>% 
  group_by(gender) %>%
  summarise(
    count = n(),
    average_gender_trip = mean(tripduration),
    sd_gender_trip = sd(tripduration)
  )

**suggestion:** In the top 3 most common station-to-station trips, it wasn't totally necessary to add gender into the arrange function before you grouped by gender since you're grouping by gender after! Normally, I would group by gender first, then try to arrange the counts after I've grouped it, then slice the head of each group. But the answer is correct in either case!

**reply:** TBH at that time I just tried different ways & this one stuck so kept it that way, but now I am updating it according to your suggestion. **Please let me know if this is what you wanted**
updated code: 
trips %>% 
  group_by(gender, start_station_name, end_station_name) %>%
  summarise(count = n()) %>%
  group_by(gender) %>%
  arrange(desc(count)) %>%
  slice_head(n=3)

PR #326, opened 2024-06-05
citibike hw

fabio-pecora:


fabio-pecora:
1) trips%>%  summarize(count = n())
2,3) trips %>%
  summarise(
    min(birth_year, na.rm = TRUE),
    max(birth_year, na.rm = TRUE)
  )

For some reason even with na.rm=TRUE I still get a null value for the min.
Do you know why?
Fabio

jennyw23:
Yep! Those look great! The "\\N" is a string that is actually not defined as an NA type in R when you first pass it in. You have a few options here: you can try to ignore the "\\N" values by ignoring/filtering them out before you try to get the min and max, or another option is to use a command called `na_if` (you can look at what it does if you're interested)

PR #327, opened 2024-06-05
citibikes in r and bash

chukwuebukad:


jennyw23:
Good work, Buka! For the first few questions, think about how you can use the pipes to answer the earliest and latest birth year questions. 

I ran into an issue testing the second to last question. I believe the issues are happening in summarize (what is the goal of adding date in the summarize function?) and then the arrange function (arrange sorts from lowest to highest)!

`trips %>%
  mutate(date = as.Date(starttime)) %>%
  group_by(date) %>%
  summarize( count = n(), date) %>%
  arrange( count ) %>%
  slice(1:1) %>%
  View`


And I'm curious as to where you got the 24 number to get the average! Hint: try out `n_distinct` to see how many distinct days there are in February (hint 2: it's not a leap year...)
trips %>%
  mutate(hour = hour(starttime)) %>%
  group_by(hour) %>%
  summarize(count = n(), count/24) %>%
  View

chukwuebukad:
I added the date to summarize because I wanted to show the date next to its corresponding count of trips. Im just realizing I should have looked for the max as well.

As for 24 I, for some reason I used the number of hours instead of number of days, I changed it to 28 before I left. I assume using n_distinct helps in taking care of days that a bike wasnt used?

jennyw23:
sounds reasonable! And yep, n_distinct takes care of days that a bike wasn't used. In our case, they were used on all 28 days of feb :)

PR #328, opened 2024-06-05
answers to citibike.R

phuongjanele:


jennyw23:
Good work overall! One note on the grep statement: `^Broadway`--the `^` symbol gets only start_street_names that _start_ with Broadway. If you want to get street names containing Broadway anywhere in the line, you'd need to remove the `^`. I think you're missing the average column in the question that asks you to count the number of trips by gender, the average trip time by gender, and the standard deviation in trip time by gender. However, the code for the standard deviation is well-written so I'm guessing you just missed that part!

PR #329, opened 2024-06-05
Today's work

suvedeisoyol:
Sorry I am requesting bunch of files, there should be one from today's assignment


jennyw23:
Good job! In the question about birth years, you can use an even simpler/more human-readable function in `mutate(across(birth_year,na_if,"\\N"))`. I'd recommend looking at `na_if` if you have time! Otherwise whoo!

PR #331, opened 2024-06-06
Day 3 Submission

Smriti9176:


jennyw23:
in this question, you only have to groupby starttime; grouping by starttime and stoptime might have issues if they are biking for multiple days!
`trips %>% 
  group_by(as_date(starttime), as_date(stoptime)) %>% 
  summarize(count=n()) %>% 
  arrange(count) %>% tail(1)`

Good job overall!

PR #332, opened 2024-06-06
Completed Plotting Problems

Songblade:


jennyw23:
Good job Shimmy! I'm going to paste the comments I added to Eric's `Good (fast) work! The plots are constructed correctly! For question 2, another way to plot the distribution by ride time might be to use a facet wrap to visualize the two usertypes separately (hard to see the customer when subscriber is overlayed on top)! another way could be to make it more transparent. Another stylistic point for the last question could be to use facet wrap to visualize all 7 days more clearly in 7 different plots--it can be hard to see what's happening in a single plot of 7 days! For the question asking for geom_smooth, it wasn't specified clearly in the instructions, but the goal is to get a straight line (linear regression!) that fits the data. What would you add inside geom_smooth to fit that?`

Another fun thing to think about is to do some zooming in / out on graphs that are really lopsided. For instance,

`trips %>%
  mutate(age = 2019 - birth_year) %>%
  group_by(age, gender) %>%
  summarize(num_trips = n()) %>%
  ggplot(aes(x = age, y = num_trips, color = gender)) +
  geom_point()
`

trimming this graph would make the data a lot easier to see!

PR #333, opened 2024-06-06
Attempt for plot_trips.R assignment

ericguan04:
Attempt for plot_trips.R assignment


jennyw23:
Good (fast) work! The plots are constructed correctly! For question 2, another way to plot the distribution by ride time might be to use a facet wrap to visualize the two usertypes separately (hard to see the customer when subscriber is overlayed on top)! another way could be to make it more transparent. Another stylistic point for the last question could be to use facet wrap to visualize all 7 days more clearly in 7 different plots--it can be hard to see what's happening in a single plot of 7 days! For the question asking for `geom_smooth`, it wasn't specified clearly in the instructions, but the goal is to get a straight line (linear regression!) that fits the data. What would you add inside `geom_smooth` to fit that?

Amazing work overall!

PR #334, opened 2024-06-06
Citibike plots

shmandelbaum:


jennyw23:
Good work! For the second question, try using facet wrap to observe the two usertypes on separate graphs! This'll be more human readable, so that customer isn't mostly covered by subscriber. Also, a tiny rec, try using fewer bins! In some cases, you might need to do 100 or a big number, but in this case, the bars become very close together and almost like a density plot! A number like 20 or 30 gets a relatively clean distribution (but it's purely stylistic). In the trips/weather section, a really clean way of doing `  mutate(sub_prcp = if_else(prcp >= .59, "T", "F")) |> ` is to simply   `mutate(rainy_day = prcp > .2) %>%`. But both get the same job done!

In the compute avg and sd question, the goal was to use `geom_ribbon` to get the standard deviation around the average line. I think I pasted an example, but let me know if you can't see it! 

image

PR #335, opened 2024-06-06
plot_trips.R solutions 

salma-ghazii:


jennyw23:
Hi Salma! Good work! Based on what Jake said today, I'd recommend adding the `aes` into the global `ggplot` just to keep it consistent (just a stylistic nit pick!)

When plotting the total number of days, you usually need to calculate the total number of trips on each day! This means transforming the dataset into a format that has a column by day and a column by total number of trips by day. This is the same idea as using a histogram (which buckets each x value into counts), which is a clever solution, but normally you'd want to be able to reference a dataset that can tell you what the highest count means! Or visualize it using dots or a different format. Without computing the total number of trips, you can't make those cool transformations.

Code looks good!

PR #336, opened 2024-06-06
day 4 work

ubaydulsami:


jennyw23:
Hi Sami! Instead of putting the `aes` function into the geom_histogram or geom_point call, putting it in `ggplot` is cleaner and more readable. In the questions about rider type, it might make more sense to use  `facet_wrap(~ usertype, scale = "free_y")` to see the distributions of both user types more easily. When plotting the total number of days, you usually need to calculate the total number of trips on each day! This means transforming the dataset into a format that has a column by day and a column by total number of trips by day. This is the same idea as using a histogram (which buckets each x value into counts), which is a clever solution, but normally you'd want to be able to reference a dataset that can tell you what the highest count means! Or visualize it using dots or a different format. Without computing the total number of trips, you can't make those cool transformations. Great work on the weather / trip questions! For the last question, you might also want to observe the mean and sd on the same graphs (i think you plotted only the value of the standard deviation instead of the average! the sd only matters in relation to the average, so you could have plotted the average with a geom ribbon using the sd.

jennyw23:
good work!

PR #337, opened 2024-06-06
plot_trips 

YinoXu:


jennyw23:
Wow, way to go above and beyond with titles and labels! For the age question, you can get the age by subtracting the year from date of birth! Basically the same idea, but your graph will look flipped. For a few questions, you use `geom_line` where it might make more sense to use a scatter plot e.g. `geom_point` to make it easier for the person looking at it to read. Sometimes `geom_line` can look very jagged. Similar idea in the trips by weather questions -- in this case the hours of a day are continuous, so you'd probably want to graph a continue line rather than a bar chart! Great work getting error bars though :) (hint: use `geom_ribbon` to get error bars in a continuous graph!)

Adding identity to the `geom_histogram` makes the graph very, very different (rider type question). When you remove it, does it look very different? Look into why this might be!

PR #338, opened 2024-06-06
answers to plot_trips.R

phuongjanele:


jennyw23:
Good work Jane! In the first problem, remember that the filter carrot in this direction, `>` asks for tripduration data more than 60, while `<` asks for data less than 60. Also, tripduration is in seconds (not minutes :)). Additionally, be careful with the position parameter in `geom_histogram`. Here's a short tutorial explaining position: https://campus.datacamp.com/courses/introduction-to-data-visualization-with-ggplot2/geometries?ex=8

After what Jake went over today, try out the weather problems when you get the chance! It's much easier to work backwards from the exact data table you need to plot (min_temp, total_trips, date). For trips_with_weather, it's not necessary to group by the precipitation variable to be able to differentiate between the low and high precipitation days. Since you've created a variable identifying precipitation, you can just add `color=precipitation` to compare the colors: `  mutate(rainy_day = prcp > .2) %>%
  ggplot(aes(x= min_temp, y = num_trips, color = rainy_day)) + 
  `

We can talk more about this today later and go through it together (it might be hard to explain in the comments!)

PR #339, opened 2024-06-07
Day 4 exercises

fabio-pecora:


jennyw23:
Good job! I was having some trouble finding the DATE and TMIN variables for the weather questions, but the plots look normal if I substitute the `ymd` and `tmin` variables. I think that affected the next plot (the columns for tmax and tmin didn't match columns 10 and 11), so just make sure to save code that changes any dataframes in the R script! 

For the trips/weather questions, I was surprised to see so few days with substantial precipitation! I see that you used the mean as the value for precipitation (which is equal to 0.09 inches). In NYC, we could probably argue that there are more days of no rain than days of rain. Looking at your if-else statement for substantial_precipitation, I wonder if those blue dots are only picking up winter days with both rain and snow... 

In compute the average number of trips and standard deviation, you're getting the mean of trip duration instead of the number of trips! Think about how this would affect the data you need to plot the avg/sd :)

jennyw23:
There were some open ended questions in Day 5 but I didn't see answers in the file. Look back at the questions and make sure you can answer them, just for your own understanding (you don't have to resubmit)

PR #340, opened 2024-06-07
Plotsssssssss

MattMunoz:


jennyw23:
Good work! For the question with usertype, try out a facet wrap to separate the two tables (or use alpha to make them transparent so you can look at the tables together)! Overall, looks great :)

PR #346, opened 2024-06-07
Joining and reshaping, diamond markdown problems

shmandelbaum:


jennyw23:
Good job! The idea of the first exercise is to show that you can get the same table of country, population, cases, and rates using either Table 2 on its own, or Table 4a+4b! The wording is a bit weird (makes you think you need to extract column by column)

PR #347, opened 2024-06-07
Attempt for Day 5 Assignments

ericguan04:
Attempt for Day 5 Assignments


jennyw23:
Good work! The re-joined table from Section 12.3.3 exercise 1 should have had the same columns when you use pivot wider and longer together. The question is hoping that you notice the data types under each column name (do you notice if stocks has the same or different datatypes?)

PR #348, opened 2024-06-07
updates for day 5

phuongjanele:


jennyw23:
It's true that the columns are in a different order doe 12.2.1, but that doesn't necessarily mean that you've lost any information! Hint: look at the datatypes in stocks vs. the rejoined table--are they the same as the original?

PR #349, opened 2024-06-07
trips and weather in R

chukwuebukad:


jennyw23:
Updated plots for day 4 look good -- it doesn't look like Day 5 exercises were committed, but that's fine as long as the assignment made sense!

PR #352, opened 2024-06-17
added the citibike test data question (Citibike_Test_Split.R)

salma-ghazii:


salma-ghazii:
Added movielens.Rmd

jennyw23:
For `# plot distribution of user activity (= number of ratings the user made)`, the histogram should be focusing on # of ratings on the x axis, where each bar represents the count of users who have made that number of ratings (rather than grouping by user id, which are just unique ids that we can't plot continuously!)

The plot for 2a looks great!

PR #353, opened 2024-06-17
week 3 day 1 exercise 

fabio-pecora:
dividing training data to 80% train, 10% test and 10% validation


fabio-pecora:
movielens exercise update

jennyw23:
commenting the same thing over from Salma's! :
For # plot distribution of user activity (= number of ratings the user made), the histogram should be focusing on # of ratings on the x axis, where each bar represents the count of users who have made that number of ratings (rather than grouping by user id, which are just unique ids that we can't plot continuously!)

The plot for 2a looks great!

PR #354, opened 2024-06-17
Week 3 day 1 & 2 work

ubaydulsami:
week 3 day 1 & 2 work submitting


jennyw23:
`# plot distribution of movie popularity (= number of ratings the movie received)` --> when visualizing a histogram of popularity, you want to look at how count of movies varies by the number of ratings (id shouldn't play a role in the counts of popularity!)

Same for user activity! User id isn't providing information about a user's activity, it should be based on the number of movies they have rated.

Graph for 2A looks great!

PR #358, opened 2024-06-17
fitting test data exercise

phuongjanele:


jennyw23:
Great work overall! The solution for plot 2A looks as expected, but we can go over the steps again in class this afternoon if it helps to think about the x and y-axes on paper!

PR #360, opened 2024-06-18
plotting movie data figures and regression from yesterday

chukwuebukad:


jennyw23:
Hi Buka! your solutions look great! I know we didn't have a lot of time to talk about plot 2a at the end of the day on Tuesday, so let's try to write it out on paper sometime this afternoon!

PR #361, opened 2024-06-18
Attempt for anatomy of a long tail assignment

ericguan04:
Attempt for anatomy of a long tail assignment + complexity_control.R assignment from yesterday
Latest commit added dashed lines to replicated graphs for the long-tail assignment


jennyw23:
`# plot distribution of movie popularity (= number of ratings the movie received)` --> think about whether movie ID on the x axis tells you anything about movie popularity. Same for the user activity question! Does user id tell you something about activity?

Plot for 2A looks great!

PR #363, opened 2024-06-18
Open Commit

MattMunoz:
The Long Tail Assingment


jennyw23:
# plot distribution of movie popularity (= number of ratings the movie received) --> think about whether movie ID on the x axis tells you anything about movie popularity. Same for the user activity question! Does user id tell you something about activity?

Plot for 2A looks great!

PR #365, opened 2024-06-20
Completed ngrams Paper

Songblade:
We have a couple of questions. Is it a problem that we did the half-life in one go? You wanted a multistage process which we didn't notice until after finishing it. What do you mean that you want half-life from 1850 to 2012? The data wouldn't get a half-life for 2012, and earlier you said you wanted it from 1875 to 1975. Finally, for smooth, do you want line or loess? The paper used loess, but smooth seems more relevant for actually evaluating their hypothesis that half-life is decreasing.


Songblade:
After a lot of installations, I also got the makefile to work (though Fabio's isn't working for reasons we can't figure out).

Songblade:
I finished my model.

PR #366, opened 2024-06-20
Added 05 final report rmarkdown and html files 

salma-ghazii:


salma-ghazii:
committed my best model rdata file

salma-ghazii:
added my 2015 data predictions in rmd file

chotomirza:
1 to 4 - not present

3a - the graph gets smoothed out because of the summarize function. Ideally, we would like to keep all the points.

Part D - Overall, the idea is correct. The graph is still a bit smoothed out. It's a good idea to set a limit in the x-axis to make the result easier to view.

Everything else looks fine!

PR #367, opened 2024-06-20
Ngram assignment

shmandelbaum:


shmandelbaum:
Added citibike_predict files

shmandelbaum:
Added Citibike 2015 Test

chotomirza:
1 to 4 - not present

3a - the graph gets smoothed out because of the summarize function. Ideally, we would like to keep all the points.

Part D - Overall, the idea is correct. The graph is still a bit smoothed out. It's a good idea to set a limit in the x-axis to make the result easier to view.

Part F: Correct, setting an x-limit here would be nice as well.

Everything else looks fine!

PR #368, opened 2024-06-20
ngrams folder updated

YinoXu:


YinoXu:
[Add best model and save it](https://github.com/msr-ds3/coursework/pull/368/commits/bfc728a1dcb22d6c4b967478046258ceabde7f4c)

chotomirza:
1 to 3 - looks good
4 - Does not have to be this complicated. All you ned to do is translate the tab characters into newlines and then filter out the lines that start with '1' or '2' from the file. You can just use the 'tr' function'

The HTML file is missing. Could you please knit the .rmd file as HTML and send it over?

YinoXu:
added the html file

PR #400, opened 2025-05-29
citibike.R and citibike.sh 

DereckTav:


chotomirza:
for the bash script ->
you can directly do: cut -d, -f# filename instead of cat filename | cut -d, -f#

cut extracts that piece of information while cat concatenates it and then you cut it (just an added step)

Other than that, everything looks good!

PR #405, opened 2025-06-16
Citibike model --Ahmed 

ahmedlotfi2510:


week4/test_citibike_predictions.Rmd
ahmedlotfi2510, line 1: # My RMSE for the 10% test in my data is 5444.888

PR #411, opened 2025-06-17
linear model

SofCora:
model is in best_model.RData


SofCora:
now i tested it with 2015 data got rmse of 32 million

PR #421, opened 2026-05-26
new assignemnt update

faridabalogun:
worked on git set up and created file as per instructions

*assignment


hannahcha417:
looks good! could you actually make it so that your txt file is your name in all lowercase?

You can just change the file name in your local repository, and then add, commit, and push the changes. It should automatically update here and I can take a look again!

hannahcha417:
looks great!

PR #427, opened 2026-05-27
Second day assignment

addinar:
Completed citibike.sh


hannahcha417:
Overall, great job! Super impressed in how you were able to pick up on command line very quickly!

A small note more generally (which I should have specified earlier), for any files ending in `.sh`, you want to include anything that's not a runnable command in a comment (by adding a '#' in front of that line). An easy way to do this is to select the lines of code that aren't comments, and press "Ctrl+ /": this turns anything selected into a comment. No need to to change anything here, but just wanted to let you know!

PR #428, opened 2026-05-27
Uploaded Nava Nizard's work for Day 2

navanizard:


hannahcha417:
Looks great; great job with learning the command line, and catching up on the previous day's assignment very quickly! 

A small note more generally (which I should have specified earlier), for any files ending in .sh, you want to include anything that's not a runnable command in a comment (by adding a '#' in front of that line). An easy way to do this is to select the lines of code that aren't comments, and press "Ctrl+ /": this turns anything selected into a comment. No need to to change anything here, but just wanted to let you know!

PR #431, opened 2026-05-27
command line exercises

AlexandruSoroiu:
day 2 exercise

### DIFF
--- a/week1/citibike.sh
+++ b/week1/citibike.sh
@@ -4,20 +4,20 @@
 #
 
 # count the number of unique stations
-
+$ cut -d, -f5 201402-citibike-tripdata.csv | sort | uniq | wc -l # 330
 # count the number of unique bikes
-
+$ cut -d, -f12 201402-citibike-tripdata.csv | sort | uniq | wc -l # 5700
 # count the number of trips per day
-
+$ cut -d, -f2 201402-citibike-tripdata.csv | cut -d ' ' -f1 | sort | uniq -c
 # find the day with the most rides
-
+$ cut -d, -f2 201402-citibike-tripdata.csv | cut -d ' ' -f1 | sort | uniq -c | sort -f -r # February 2nd
 # find the day with the fewest rides
-
+$ cut -d, -f2 201402-citibike-tripdata.csv | cut -d ' ' -f1 | sort | uniq -c | sort -f # February 13th
 # find the id of the bike with the most rides
-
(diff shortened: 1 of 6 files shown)


week1/citibike.sh
```diff
@@ -4,20 +4,20 @@
...
+$ cut -d, -f12 201402-citibike-tripdata.csv | sort | uniq -c | sort -f -r | head -n1 # 130 entries with the bike id of 20837
 # count the number of rides by gender and birth year
-
+$ cut -d, -f14,15 201402-citibike-tripdata.csv | sort | uniq -c # wc -l is 142
 # count the number of trips that start on cross streets that both contain numbers (e.g., "1 Ave & E 15 St", "E 39 St & 2 Ave", ...)
-
-
+$ cut -d, -f5 201402-citibike-tripdata.csv | grep ".*[0-9].*&.*[0-9].*" # wc -l is 90549
 # compute the average trip duration
+$ awk -F, 'BEGIN{duration=0; total=0} {duration += $1;total++} END{print duration/total}' 201402-citibike-tripdata.csv # 874.516
```
hannahcha417, line 23: A small note that you don't necessarily need to initialize `duration` and `total` for awk, but this is good practice more generally for other programming languages!

week1/citibike.sh
```diff
@@ -4,20 +4,20 @@
...
+$ cut -d, -f2 201402-citibike-tripdata.csv | cut -d ' ' -f1 | sort | uniq -c
 # find the day with the most rides
-
+$ cut -d, -f2 201402-citibike-tripdata.csv | cut -d ' ' -f1 | sort | uniq -c | sort -f -r # February 2nd
 # find the day with the fewest rides
-
+$ cut -d, -f2 201402-citibike-tripdata.csv | cut -d ' ' -f1 | sort | uniq -c | sort -f # February 13th
 # find the id of the bike with the most rides
-
+$ cut -d, -f12 201402-citibike-tripdata.csv | sort | uniq -c | sort -f -r | head -n1 # 130 entries with the bike id of 20837
```
hannahcha417, line 17: Just a note here that you don't need the `-f` flag for `sort`; the `-f` flag is essentially for making the sorting comparison case insensitive (e.g. making "apple" and "Apple" be considered the same). Since you're dealing with numeric IDs for the bikes, this shouldn't be an issue for this case specifically!

hannahcha417:
Great job at picking up on command line super quickly! :)

I left a few small comments, but all your commands are correct! A small note that you don't need the `$` character in front of all your commands in the assignment file (no need to fix, but just something to know for working in `.sh` files in the future); you can just directly paste the command in as you would write it in your terminal!

PR #432, opened 2026-05-27
Day 2 assignment

ChanaRapp:
This is my edited, runable submission - thank you!

### DIFF
--- a/week1/citibike.sh
+++ b/week1/citibike.sh
@@ -3,16 +3,18 @@
 # add your solution after each of the 10 comments below
 #
 
+# I tried to make all my solutions output exactly the information that was asked for - no more
+
 # count the number of unique stations
-# cut -d, -f4,8 201402-citibike-tripdata.csv | tr , '\n' | sort | uniq | wc -l
+cut -d, -f4,8 201402-citibike-tripdata.csv | tr , '\n' | sort | uniq | wc -l
 # Answer: 331 (-2 for headers)
 
 # count the number of unique bikes
-# cut -d, -f12 201402-citibike-tripdata.csv | sort | uniq | wc -l 
+cut -d, -f12 201402-citibike-tripdata.csv | sort | uniq | wc -l 
 # Answer: 5700
 
 # count the number of trips per day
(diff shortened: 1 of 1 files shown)


week1/citibike.sh
```diff
@@ -3,21 +3,68 @@
 # add your solution after each of the 10 comments below
 #
 
+# I tried to make all my solutions output exactly the information that was asked for - no more
+
 # count the number of unique stations
+cut -d, -f4,8 201402-citibike-tripdata.csv | tr , '\n' | sort | uniq | wc -l
+# Answer: 331 (-2 for headers)
```
hannahcha417, line 10: Since you did `uniq`, it should only be -1 for the header, I think!

ChanaRapp:
Thank you for the feedback! I thought it might be -2 because I split each
line into 2 with 'tr , '\n'.

week1/citibike.sh
```diff
@@ -3,21 +3,68 @@
...
 # count the number of rides by gender and birth year
+cut -d, -f14,15 201402-citibike-tripdata.csv | sort -bn | uniq -c
 
 # count the number of trips that start on cross streets that both contain numbers (e.g., "1 Ave & E 15 St", "E 39 St & 2 Ave", ...)
+cut -d, -f5 201402-citibike-tripdata.csv | grep '[0-9].*&.*[0-9]' | wc -l
+# Answer: 90549
 
 
 # compute the average trip duration
+awk -F, '1 ~ /[0-9]*/{counts[0]++; duration[0]+=$1} END {print(duration[0]/counts[0])}' 201402-citibike-tripdata.csv
```
hannahcha417, line 69: For this part of your code: `1 ~ /[0-9]*/`, what you're essentially doing is asking if the literal number 1 matches the pattern `/[0-9]*/`, which is always true! I think you meant to include a $ in front of the 1, and also instead of ` /[0-9]*/`, ` /[0-9]+/`, because the first one is essentially asking if it is zero or more digits, which wouldn't actually filter out non-digit rows!

Overall, though, your command works without this filtering since we have well-formed data! But it's great that you're thinking about this in this way!

hannahcha417:
Overall, great work on the assignment; impressed with how you were able to pick up on command line!

I left a few notes of feedback, but you're all good in terms of correctness. Again, great work! :)

PR #433, opened 2026-05-27
Command Line Assignment

faridabalogun:
Assignment 2 - Command line assignment

### DIFF
--- a/week1/citibike.sh
+++ b/week1/citibike.sh
@@ -4,20 +4,28 @@
 #
 
 # count the number of unique stations
-
+cut -d, -f5 201402-citibike-tripdata.csv | sort | uniq | wc -l
+#330
 # count the number of unique bikes
-
+cut -d, -f12 201402-citibike-tripdata.csv | sort | uniq | sort -n | wc -l
+#5700
 # count the number of trips per day
-
-# find the day with the most rides
-
+cut -d, -f2 201402-citibike-tripdata.csv |cut -d" " -f1 | sort | uniq -c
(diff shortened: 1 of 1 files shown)


week1/citibike.sh
```diff
@@ -4,20 +4,28 @@
...
-
+cut -d, -f2 201402-citibike-tripdata.csv |cut -d" " -f1 | sort | uniq -c | sort -n|head -n2
+#876
 # find the id of the bike with the most rides
-
+cut -d, -f12 201402-citibike-tripdata.csv | sort | uniq -c | sort -nr|head -n1
+#20837
 # count the number of rides by gender and birth year
-
+cut -d, -f14,15 201402-citibike-tripdata.csv | sort | uniq -c| sort -n|wc -l
```
hannahcha417, line 24: Just a note here that you're returning the number of unique gender and birth year combinations exist in the data set! We didn't necessarily specify that we wanted the number of unique combinations, so our solution was just a subset of your full command: `cut -d, -f14,15 201402-citibike-tripdata.csv | sort | uniq -c`

Nothing wrong with this, but just wanted to note exactly what you were returning! Also, if you just want to return line count in the future (through `wc -1`), you don't necessarily need to sort beforehand, since sorting won't change how many total values there are!

hannahcha417:
Overall, great work! It's great how quickly you were able to pick up on command line :)

I left a small comment on one of your commands (nothing to change, just wanted to point out a few things), but again, good job!

PR #434, opened 2026-05-27
Day Two Assignment

alexie-louise:
LINUX Shell Scripting Practice Exercises with CitiBike Data


week1/citibike.sh
```diff
@@ -4,20 +4,223 @@
 #
 
 # count the number of unique stations
+# Input: 
+cut -d, -f5,9 201402-citibike-tripdata.csv | uniq | wc -l
```
hannahcha417, line 8: You're on the right track! But essentially what you're doing here is counting the number of unique beginning x end station combinations. You can just choose beginning or end station, and just count the unique values for just one of the columns, not the combination of both!

week1/citibike.sh
```diff
@@ -4,20 +4,223 @@
...
 
+# Output: 223435
 # count the number of unique bikes
+# Input:
+cut -d, -f12 201402-citibike-tripdata.csv | uniq | wc -l
```
hannahcha417, line 13: This is close! You're actually going to want to `sort` before you run `uniq`, because `uniq` is a bit quirky. It essentially only compares the value to the one before it, rather than all the existing values before. So, it tends to overestimate quite a bit if you don't `sort` before!

week1/citibike.sh
```diff
@@ -4,20 +4,223 @@
...
+     11 1901,1
+     68 1900,1
+      9 1899,1
+      1 birth year,gender
+   6717 \N,0
+"""
 # count the number of trips that start on cross streets that both contain numbers (e.g., "1 Ave & E 15 St", "E 39 St & 2 Ave", ...)
-
+#Input: 
+cut -d, -f5 201402-citibike-tripdata.csv | grep '& [0-9]' | wc -l
```
hannahcha417, line 219: This is close! Currently, your grep call only checks that there's a number after the "&" sign, which applies to any combination where the second location with a cross street, Instead, you want to check both sides of the "&" sign. So, you would want the number pattern checking to be on the left side of the expression as well.

You should also pad everything with `.*`, which allows there to be any character to be in the location you put it. For instance, without it, right now you're checking that the character directly following the "&" is a number, which would not be able to catch "[...] & E 15 St", for example

hannahcha417:
Overall, great job with the assignment! You've shown a great understanding of command line :) Most of your problem solutions are correct, but I did find a few places that have an incorrect answer (but can easily be fixed, since you were very close!) I left a few comments for those cases that I hope can be helpful, but feel free to message me if you have any questions!

PR #436, opened 2026-05-28
day 2 assignment

pennywa:


week1/citibike.sh
```diff
@@ -4,20 +4,44 @@
...
+#...
+#2014-02-28 9587
+#starttime 1
 
 # find the day with the most rides
+cut -d, -f2 201402-citibike-tripdata.csv | cut -d' ' -f1 | sort | uniq -c | sort -nr | head -1 | awk '{print $2, $1}'
+#2014-02-02 13816
 
 # find the day with the fewest rides
+tail -n +2 201402-citibike-tripdata.csv | cut -d, -f2 | cut -d' ' -f1 | sort | uniq -c | sort -n | head -1 | awk '{print $2, $1}'
```
hannahcha417, line 26: I saw your comment below about how having cut before doesn't work, but the below command returns the same as above:
`cut -d, -f2 201402-citibike-tripdata.csv | tail -n +2 | cut -d' ' -f1 | sort | uniq -c | sort -n | head -1 | awk '{print $2, $1}'`

Let me know if there's something specific you were trying to do that wasn't working!

hannahcha417:
Overall, great work! Super impressed with how quickly you were able to pick up on command line, and appreciate the extra work you put in to filter out the header rows!

I left a comment on a question you had in the comments for that question, but let me know if you need an additional clarification on anything!

PR #438, opened 2026-05-28
Uploaded Nava Nizard's work for Day 3

navanizard:


week1/citibike.R
```diff
@@ -23,26 +23,77 @@ trips <- mutate(trips, gender = factor(gender, levels=c(0,1,2), labels = c("Unkn
...
+    filter(grepl('Broadway', start_station_name) 
+        | grepl('Broadway', end_station_name))
 
 # do the same, but find all trips that both start and end on broadway
+trips %>% 
+    filter(grepl('Broadway', start_station_name) 
+        & grepl('Broadway', end_station_name))
 
 # find all unique station names
+c(pull(trips, start_station_name, end_station_name)) %>% summary(n_distinct)
```
hannahcha417: Actually, for `pull`, the third argument is name, which accepts a column to use as the names of the resulting vector. So, what you're doing here is returning a named vector where the values are `start_station_name`, and the names are `end_station_name`. You do return the correct number though, since both start and end stations have 329 unique stations! Also, `n_distinct` can't be passed as an argument to `summary()`; so your current return is just the return of `summary()` itself!

If you wanted to count the distinct station names for both start and end, you would have to concatenate both columns into one vector and count the unique values. You could do something like: `c(trips$start_station_name, trips$end_station_name) %>% n_distinct()`

week1/citibike.R
```diff
@@ -23,26 +23,77 @@ trips <- mutate(trips, gender = factor(gender, levels=c(0,1,2), labels = c("Unkn
...
+    summarize(count = n()) %>% 
+    arrange(desc(count)) %>%
+    head(10)
 
 # find the top 3 end stations for trips starting from each start station
+trips %>% 
+    group_by(start_station_name) %>% 
+    mutate(rank = row_number()) %>% 
+    filter(rank <= 3) %>% 
+    select(end_station_name)
```
hannahcha417, line 71: This is close! But actually ` mutate(rank = row_number())` just assigns a row number in each group, but it doesn't rank by the frequency of the end stations. So, your filter call is just picking the first 3 rows for each start station, regardless of the frequency of end stations.

You're going to want to do a similar thing to the previous answer, where you group by both `start_station_name` and `end_station_name`, then get the counts of each pair, and group by start station, then figure out how to get the top 3 end stations for each!

week1/citibike.R
```diff
@@ -23,26 +23,77 @@ trips <- mutate(trips, gender = factor(gender, levels=c(0,1,2), labels = c("Unkn
...
 
 # find the top 3 most common station-to-station trips by gender
+trips %>%
+    group_by(gender) %>%
+    mutate(rank = row_number()) %>%
+    filter(rank <= 3)
```
hannahcha417: Same idea here; currently, you're not ordering them in the way that you would like! You're going to want to group by gender, start_station_name, and end_station name. Then, you can get the counts, order them by frequency, and then get the top 3 stations!

hannahcha417:
Overall, great job with the assignment! Most of your commands were correct, and I left a few comments in places that could be improved/corrected!

week1/citibike.R
navanizard: Thanks! I'll fix that now.

PR #439, opened 2026-05-28
Third Day Assignment

addinar:
Completed citibike.R


week1/citibike.R
```diff
@@ -23,26 +23,105 @@ trips <- mutate(trips, gender = factor(gender, levels=c(0,1,2), labels = c("Unkn
...
+    head(n=1)
+# ans: 1899
+
+trips %>% 
+    select(birth_year) %>% 
+    arrange(desc(birth_year)) %>% 
+    filter(grepl("[0-9]", birth_year)) %>% 
+    head(n=1)
+# ans: 1997
+
```
hannahcha417, line 45: This is great! An easier way to do this as well would be to replace the "\\N" in the birth_year column with "NA" through ` mutate(birth_year = na_if(birth_year, "\\N")) `, and then use the `na.rm` argument in `min()` and `max()`. For instance, you could do `summarize(min_birth_year = min(birth_year, na.rm = T), max_birth_year = max(birth_year, na.rm = T))`; `na.rm = T` (T for True) removes "NA" values and without it, min and max would return NA!

week1/citibike.R
```diff
@@ -23,26 +23,105 @@ trips <- mutate(trips, gender = factor(gender, levels=c(0,1,2), labels = c("Unkn
...
+    head(n=10) %>%
+    select(start_station_name, end_station_name, frequency) 
 
 # find the top 3 end stations for trips starting from each start station
+trips %>%
+    group_by(start_station_name, end_station_name) %>%
+    summarize(frequency=n()) %>%
+    arrange(start_station_name, desc(frequency)) %>%
+    mutate(rank=row_number()) %>%
+    filter(rank<=3)
```
hannahcha417, line 91: Instead of the last two lines, you could also just do `slice(1:3)`!

week1/citibike.R
```diff
@@ -23,26 +23,105 @@ trips <- mutate(trips, gender = factor(gender, levels=c(0,1,2), labels = c("Unkn
...
 # compute the average number of trips taken during each of the 24 hours of the day across the entire month
+trips %>%
+   mutate(hour=hour(starttime)) %>%
+   group_by(hour) %>%
+   summarize(frequency=n(), num_dates=n_distinct(as.Date(starttime))) %>%
+   mutate(avg_frequency=frequency/num_dates) %>%
+   print(n = 24)
+
 # what time(s) of day tend to be peak hour(s)?
+# Peak hours are around 4pm-6pm
```
hannahcha417, line 120: Correct! Intuitively, this makes sense

hannahcha417:
Great job with this assignment!! All your commands returned correct output; I just left a few small suggestions!

PR #441, opened 2026-05-28
Day 3 assignment

ChanaRapp:
Here is my completed work. Thank you!


week1/citibike.R
```diff
@@ -22,27 +22,77 @@ trips <- mutate(trips, gender = factor(gender, levels=c(0,1,2), labels = c("Unkn
...
+            sd_trip_duration = sd(tripduration))
+
+# 7. find the 10 most frequent station-to-station trips
+trips %>% 
+    select(start_station_name, end_station_name) %>% 
+    group_by(start_station_name, end_station_name) %>%
+    summarize(count = n()) %>%
+    arrange(desc(count)) %>%
+    head(n=10) %>%
+    select(start_station_name, end_station_name)
```
hannahcha417, line 58: Technically, you can omit the `select` and have it work as intended! If you don't add the select at the end, you can also see the count column, which could be helpful :)

hannahcha417:
Great job overall! I left a small comment but all your answers were correct; well done!

PR #443, opened 2026-05-29
assignment day 3 for review

AlexandruSoroiu:


week1/citibike.R
```diff
@@ -23,26 +23,77 @@ trips <- mutate(trips, gender = factor(gender, levels=c(0,1,2), labels = c("Unkn
...
 # find the earliest and latest birth years (see help for max and min to deal with NAs)
-
+trips %>%
+    select(birth_year) %>%
+    arrange(desc(birth_year)) %>%
+    filter(grepl("[0-9]", birth_year)) %>%
+    head(n=1) # latest is 1997 and earliest is 1899 by not using "desc"
 # use filter and grepl to find all trips that either start or end on broadway
-
+select(filter(trips, grepl("Broadway", start_station_name) | grepl("Broadway", end_station_name)), start_station_name, end_station_name)
```
hannahcha417, line 34: You can actually combine these pretty cleanly like this: `trips %>%
  filter(grepl('Broadway', start_station_name) | grepl('Broadway', end_station_name))`

week1/citibike.R
```diff
@@ -23,26 +23,77 @@ trips <- mutate(trips, gender = factor(gender, levels=c(0,1,2), labels = c("Unkn
...
 # count the number of trips (= rows in the data frame)
-
+nrow(trips) # 224736
 # find the earliest and latest birth years (see help for max and min to deal with NAs)
-
+trips %>%
+    select(birth_year) %>%
+    arrange(desc(birth_year)) %>%
+    filter(grepl("[0-9]", birth_year)) %>%
+    head(n=1) # latest is 1997 and earliest is 1899 by not using "desc"
```
hannahcha417, line 32: This is a good way to filter out the "NA"s in the birth year! Another way is like so:
`mutate(birth_year = na_if(birth_year, "\\N"))`, which transforms all the "\\N" into "NA". Then, you can add the folllowing argument to min/max to drop the NAs:
  `summarize(min_birth_year = min(birth_year, na.rm = True),
            max_birth_year = max(birth_year, na.rm = True))`

week1/citibike.R
```diff
@@ -23,26 +23,77 @@ trips <- mutate(trips, gender = factor(gender, levels=c(0,1,2), labels = c("Unkn
...
+    summarize(frequency = n()) %>%
+    arrange(desc(frequency)) %>%
+    head(n = 10)
 # find the top 3 end stations for trips starting from each start station
+trips %>%
+    group_by(start_station_name, end_station_name) %>%
+    summarize(frequency = n()) %>%
+    arrange(start_station_name, desc(frequency)) %>%
+    mutate(rank=row_number()) %>%
+    filter(rank<=3)
```
hannahcha417, line 69: You can also do `slice(1:3)` instead of the last two lines!

hannahcha417:
Overall, great job on the assignment! I left a few small suggestions, but all your answers were correct. Great work!

PR #446, opened 2026-05-29
day 3 assignment

pennywa:


week1/citibike.R
```diff
@@ -23,26 +23,88 @@ trips <- mutate(trips, gender = factor(gender, levels=c(0,1,2), labels = c("Unkn
 ########################################
 
 # count the number of trips (= rows in the data frame)
+nrow(filter(trips))
```
hannahcha417, line 26: You actually don't need the filter here!

week1/citibike.R
```diff
@@ -23,26 +23,88 @@ trips <- mutate(trips, gender = factor(gender, levels=c(0,1,2), labels = c("Unkn
...
+    slice_max(trip_count,n = 3) 
 
 # find the day with the most trips
 # tip: first add a column for year/month/day without time of day (use as.Date or floor_date from the lubridate package)
+(most_trips_day <- trips) %>%
+mutate(trip_date = floor_date(starttime, unit = "day")) %>%
+group_by(trip_date) %>% 
+summarize(daily_count = n()) %>%
+slice_max(daily_count, n = 1)
+# What's the difference between using floor date and as.Date?
```
hannahcha417, line 95: I think the output format/object is slightly different, but they're essentially the same!

hannahcha417:
Overall, great job! I just left a few minor comments/suggestions, but all your answers are correct! :)

PR #452, opened 2026-06-01
Uploaded Nava Nizard's Work for Day 4

navanizard:


week1/plot_trips.R
```diff
@@ -18,25 +20,102 @@ load('trips.RData')
 ########################################
 
 # plot the distribution of trip times across all rides (compare a histogram vs. a density plot)
+trips %>% 
+    filter(tripduration <= 40000) %>% #filter out all trips longer than 24 hours
+    group_by(tripduration) %>%
+    summarise(count = n()) %>%
+    ggplot(aes(x = tripduration / 60)) +
+    geom_histogram(bins = 10) +
```
hannahcha417, line 28: You could add more bins to see the distribution better!

week1/plot_trips.R
```diff
@@ -47,16 +126,73 @@ trips_with_weather <- inner_join(trips, weather, by="ymd")
...
+    labs(
+        x = 'Minimum Temperature', 
+        y = 'Number of Trips', 
+        title = 'Number of Trips as a Function of the Minimum Temperature') 
 
 # repeat this, splitting results by whether there was substantial precipitation or not
 # you'll need to decide what constitutes "substantial precipitation" and create a new T/F column to indicate this
+trips_with_weather %>% 
+    mutate(substantial_prcp = (prcp >= mean(prcp))) %>%
+    group_by(tmin, substantial_prcp) %>%
```
hannahcha417, line 143: If you group by tmin instead of ymd, you're counting trips per temperature value instead of per day. Since different days could have the same tmin, those days would get merged together!

week1/plot_trips.R
```diff
@@ -47,16 +126,73 @@ trips_with_weather <- inner_join(trips, weather, by="ymd")
...
+        title = 'Number of Trips as a Function of the Minimum Temperature') + 
+    scale_y_continuous(label = comma)
 
 # add a smoothed fit on top of the previous plot, using geom_smooth
-
-# compute the average number of trips and standard deviation in number of trips by hour of the day
+trips_with_weather %>% 
+    mutate(substantial_prcp = (prcp >= mean(prcp))) %>%
+    group_by(tmin, substantial_prcp) %>%
+    summarize(count = n()) %>%
```
hannahcha417, line 158: Same comment as above here, you should be grouping by the day instead of tmin!

week1/plot_trips.R
```diff
@@ -47,16 +126,73 @@ trips_with_weather <- inner_join(trips, weather, by="ymd")
...
+    group_by(hr, day) %>%
+    summarize(count = n()) %>%
+    group_by(hr) %>%
+    summarize(avg = mean(count), std = sd(count)) %>%
+    ggplot(aes(x = hr, y = avg, ymin = avg - std, ymax = avg + std)) +
+    geom_pointrange() +
+    labs(
+        x = 'Hour of the Day', 
+        y = 'Average Number of Trips', 
+        title = 'Avg Number of Trips and Std in Number of Trips by Hour of the Day')
```
hannahcha417, line 183: Another way to do this could be through geom_line and geom_ribbon to show the SD (you can see the parameters for geom_ribbon in documentation)!

hannahcha417:
Overall, great work on the assignment and nice graphs! I left a few comments/suggestions; the only major comment is that for your graphs on the number of trips as a function of the minimum temperature per day, you're going to want to group_by the day, rather than the precipitation (because days can have the same precipitation)

week1/plot_trips.R
navanizard: To clarify, I would need to include ymd in the grouping in addition to the tmin and substantial_prcp, and not exclude the tmin (since I need to access those values after the group_by and summarize)?

PR #454, opened 2026-06-01
Week 1, Day 4 + Week 2, Day 1

alexie-louise:


week1/plot_trips.R
```diff
@@ -18,25 +18,80 @@ load('trips.RData')
 ########################################
 
 # plot the distribution of trip times across all rides (compare a histogram vs. a density plot)
+ggplot(trips, mapping = aes(x = tripduration)) +
+    geom_histogram(bins = 200, fill = 'red') + 
+    scale_x_log10(label = comma)
+
+ggplot(trips, mapping = aes(x = tripduration)) +
+    geom_density(fill = '#b73ab1') + 
+    scale_x_log10(label = comma)
```
hannahcha417, line 28: I would recommend filtering out unusually large values (e.g. 1 million seconds) over having to use log scale for the x axis!

week1/plot_trips.R
```diff
@@ -18,25 +18,80 @@ load('trips.RData')
...
+    geom_density() + 
+    scale_x_log10(label = comma) + 
+    facet_wrap(~ usertype)
 
 # plot the total number of trips on each day in the dataset
+trips %>%
+mutate(days = floor_date(starttime, unit = "day")) %>%
+group_by(days) %>%
+    ggplot(mapping = aes(x = days)) + 
+    geom_bar()
```
hannahcha417, line 50: This might be better represented as a scatter plot or a line graph!

week1/plot_trips.R
```diff
@@ -18,25 +18,80 @@ load('trips.RData')
...
-
+# NTS: add a bar plot
+trips %>%
+mutate(age = 2014 - birth_year) %>%
+group_by(age, gender) %>%
+summarize(count = n()) %>%
+    ggplot(mapping = aes(x = age, y = count, color = gender)) + 
+    geom_point() +
+    geom_smooth()
 # plot the ratio of male to female trips (on the y axis) by age (on the x axis)
```
hannahcha417, line 60: You could also lower the y axis to make it easier to see!

week1/plot_trips.R
```diff
@@ -47,16 +102,73 @@ trips_with_weather <- inner_join(trips, weather, by="ymd")
...
+    geom_point()
+
+trips_with_weather %>%
+mutate(sub_prcp = prcp > 0.0936) %>%
+group_by(tmin, date, sub_prcp) %>%
+summarize(trips = n()) %>%
+ggplot(mapping = aes(x = tmin, y = trips, color = sub_prcp)) +
+    geom_point() + 
+    facet_wrap(~ sub_prcp)
 # add a smoothed fit on top of the previous plot, using geom_smooth
```
hannahcha417, line 137: Since you're already separating by color on the graph, you could also just put them together on the same graph rather than doing facet wrap!

week1/plot_trips.R
```diff
@@ -47,16 +102,73 @@ trips_with_weather <- inner_join(trips, weather, by="ymd")
...
+trips_with_weather %>%
+mutate(wday = wday(starttime, label = TRUE)) %>%
+group_by(wday, date) %>%
+summarize(count = n()) %>%
+group_by(wday) %>%
+summarize(avg = mean(count), sd = sd(count)) %>%
+    ggplot(aes(x = wday, y = avg)) + 
+        geom_pointrange(aes(ymin = avg - sd, ymax = avg + sd)) +
+        xlab('Day of the Week') +
+        ylab('Avg Num of Trips')
```
hannahcha417, line 174: For this problem, I would actually recommend using facet_wrap to show each day of the week! Right now, you're essentially representing the average number of trips per day of the week, rather than the average number of trips per hour for each day of the week, which are subtly different!

hannahcha417:
Overall, great work and great graphs! I left a few comments/suggestions on things you can improve on. I'll come back to your week 2 assignment once I wrap up looking at everyone's day 4 assignment!

week2/combining-reshaping.R
```diff
@@ -43,4 +82,18 @@ people <- tribble(
...
+values_fn = list(values = mean)
+)
+
+"""
+If we attempted to widen the table, we'd pivot the names column to have the entries age and height
+as their own columns. It would take the values from the values column. However, following through
+with implementation, we run into an error. Phillip Woods has two observations of differing ages.
+This creates a data constraint because one cell cannot contain two data points. We solve this by
+passing through the values_fn argument to aggregate duplicate values.
+"""
```
hannahcha417, line 99: I would actually not take the mean of the values; they might actually be two different people with different ages that happen to have the same name! Instead, you can group_by(name,names), then add a row_number() column, then pivot_wider based on that; this would separate out cases based on unique row number

week2/diamond-sizes.Rmd
```diff
@@ -29,10 +29,25 @@ smaller %>%
...
+instead of setting echo = FALSE on each chunk, set a global option.
+knitr::opts_chunk$set(
+  echo = FALSE
+)
+```{r}
+smaller %>% 
+  ggplot(mapping = aes(x = cut)) + 
+  geom_bar(mapping = aes(fill = clarity)) +
+  facet_wrap(~ color)
+```
```
hannahcha417, line 53: No need to, but could also be nice to accompany the plot with some text explanation!

hannahcha417:
Great job on the week 2 day 1 assignment! I left a few comments here and there!

PR #455, opened 2026-06-01
Completed day 4 assignment

addinar:
Completed week1/plot_trips.R


week1/plot_trips.R
```diff
@@ -17,26 +17,132 @@ load('trips.RData')
...
+
+trips %>%
+    ggplot(aes(x=tripduration)) +
+    geom_density(fill="blue") +
+    scale_x_log10(label=comma) +
+    labs(
+        x="Trip Duration",
+        y="Count",
+        title="Disribution of Trip Times Across Rides"
+    )
```
hannahcha417, line 42: For both the above, you'll probably want to filter out unusually large values! (e.g. it's unlikely that a bike ride lasted 1million seconds). Then, you won't have to do a log scale for the x axis

week1/plot_trips.R
```diff
@@ -17,26 +17,132 @@ load('trips.RData')
...
+trips %>%
+    ggplot(aes(x=tripduration, color=usertype, fill=usertype)) + 
+    geom_density() +
+    scale_x_log10(label=comma) +
+    labs(
+        x="Trip Duration",
+        y="Count",
+        title="Disribution of Trip Times by Rider Type"
+    ) +
+    facet_wrap(~ usertype)
```
hannahcha417, line 64: You can use ` facet_wrap(~ usertype, scale = "free_y")` here since the y ranges between the two user types are very different! This will base the y axis on the graph itself, not both graphs!

week1/plot_trips.R
```diff
@@ -17,26 +17,132 @@ load('trips.RData')
...
+    group_by(date = as.Date(starttime)) %>%
+    summarize(frequency=n()) %>%
+    ggplot(aes(x=date, y=frequency)) + 
+    geom_point() + 
+    geom_smooth() + 
+    labs(
+        x="Day",
+        y="Total trips",
+        title="Total Trips per Day"
+    )
```
hannahcha417, line 77: Great job going the extra step to add the fitted line!

week1/plot_trips.R
```diff
@@ -17,26 +17,132 @@ load('trips.RData')
...
 # plot the total number of trips (on the y axis) by age (on the x axis) and gender (indicated with color)
+trips %>%
+    mutate(age=2014-birth_year) %>%
+    ggplot(aes(x=age, color=gender)) +
+    geom_bar(alpha=0.2) +
+    labs(
+        x="Age",
+        y="Total trips",
+        title="Total Trips by Age and Gender"
+    )
```
hannahcha417, line 88: This might be better represented as a scatter plot or a line graph! The opacity does help, but that could make it a bit more clear :)

week1/plot_trips.R
```diff
@@ -17,26 +17,132 @@ load('trips.RData')
...
 # hint: use the pivot_wider() function to reshape things to make it easier to compute this ratio
 # (you can skip this and come back to it tomorrow if we haven't covered pivot_wider() yet)
 
+
 ########################################
 # plot weather data
 ########################################
 # plot the minimum temperature (on the y axis) over each day (on the x axis)
+weather %>%
+    ggplot(aes(x=date, y=tmin)) +
```
hannahcha417, line 123: You should do x = ymd for a cleaner x axis!

hannahcha417:
Overall, great job and great graphs! I left a few comments/suggestions here and there that are mostly potential improvements you can make. Good work! :)

PR #456, opened 2026-06-01
Day 4 Assignment

jcqln0:
jacqueline day four assignment


week1/plot_trips.R
```diff
@@ -18,26 +18,91 @@ load('trips.RData')
...
 # plot the distribution of trip times across all rides (compare a histogram vs. a density plot)
+ggplot(trips, aes(x=tripduration)) +
+    scale_x_log10(label = comma) +
+    geom_histogram(bins = 200) + 
+    labs(x = "Trip Duration", x = "Trip Count", title = "Distribution of Trip Time Across All Rides")
+
+ggplot(trips, aes(x=tripduration)) +
+    scale_x_log10(label = comma) +
+    geom_density(fill = "#bbbbf8", colour = "#bbbbf8") +
+    labs(x = "Trip Duration", y = "Trip Count", title = "Distribution of Trip Times Across All Rides")
```
hannahcha417, line 29: For both of the above, I would recommend filtering out some of the unusually large values! For instance, it's unlikely the case that a bike ride was over 1 million seconds. You can set something reasonable yourself, but this should also make it so you don't have to use a log scale x axis!

week1/plot_trips.R
```diff
@@ -18,26 +18,91 @@ load('trips.RData')
...
 
 # plot the distribution of trip times by rider type indicated using color and fill (compare a histogram vs. a density plot)
+ggplot(trips, aes(x=tripduration, colour=usertype, fill=usertype)) +
+    scale_x_log10(label = comma) +
+    geom_histogram(bins = 50) +
+    facet_wrap(~ usertype) +
+    labs(x = "Trip Duration", y = "Trip Count", title = "Distribution of Trip Times Across All Rides (Split by User Type)")
+
```
hannahcha417, line 37: For this graph, you can do `facet_wrap(~ usertype, scale = "free_y")` to set different ranges for both graphs depending on the user type! This can help make the graphs more readable, since the range for the left is much smaller than the right

week1/plot_trips.R
```diff
@@ -18,26 +18,91 @@ load('trips.RData')
...
+    scale_x_log10(label = comma) +
+    geom_density() +
+    labs(x = "Trip Duration", y = "Density", title = "Distribution of Trip Times Across All Rides (Colored by User Type)")
 
 # plot the total number of trips on each day in the dataset
+trips %>%
+    mutate(day = floor_date(starttime, unit = "day")) %>%
+    ggplot(aes(x=day)) +
+    geom_bar(fill = '#9797c0') +
+    labs(x = "Day", y = "Trip Count", title = "Total number of trips per day")
```
hannahcha417, line 49: This could be better represented as a line graph or a scatter plot!

week1/plot_trips.R
```diff
@@ -18,26 +18,91 @@ load('trips.RData')
...
+
+# Scatter plot version (maybe it makes sense? not completely sure)
+trips %>%
+    mutate(age = 2014 - birth_year) %>%
+    group_by(age, gender) %>%
+    summarize(count = n()) %>%
+    ggplot(aes(x = age, y = count, colour = gender)) +
+    geom_point() +
+    geom_smooth() +
+    labs(x = "Age", y = "Total number of trips", title = "Total number of trips vs age (colored by gender)")
```
hannahcha417, line 65: In this case, the scatter plot version is much better, since you can see the differences between groups much more clear!

hannahcha417: You can also limit the y axis a bit to make it easier to see!

week1/plot_trips.R
```diff
@@ -18,26 +18,91 @@ load('trips.RData')
...
+    geom_point(colour = '#6d6daf') +
+    labs(x = "Age", y = "Ratio of male to female trips", title = "Age vs Ratio of male to female trips")
 
 ########################################
 # plot weather data
 ########################################
 # plot the minimum temperature (on the y axis) over each day (on the x axis)
+ggplot(weather, aes(x = date, y = tmin)) +
+    geom_point(colour='#618695') +
+    labs(x = "Date", y = "Minimum Temperature", title = "Date vs Minimum temperature")
```
hannahcha417, line 88: you should set x = ymd for a cleaner x axis!

week1/plot_trips.R
```diff
@@ -18,26 +18,91 @@ load('trips.RData')
...
+    geom_point() +
+    labs(x = "Date", y = "Temperature Ranges", title = "Date vs Min and Max Temperatures (colored by min/max)")
+
+weather %>%
+    pivot_longer(names_to = "maxmin", values_to = "ex_temp", 5:6) %>%
+    ggplot(aes(x = date, y = ex_temp, group = maxmin, colour = maxmin)) +
+    geom_line() +
+    labs(x = "Date", y = "Temperature Ranges", title = "Date vs Min and Max Temperatures (colored by min/max)")
+
+# reshaping data
```
hannahcha417, line 105: Either plot is good in this case!

week1/plot_trips.R
```diff
@@ -47,16 +112,89 @@ trips_with_weather <- inner_join(trips, weather, by="ymd")
...
+
+trips_with_weather %>% 
+    mutate(is_substantial = (prcp > 0.0936)) %>%
+    group_by(tmin, date, is_substantial) %>%
+    summarize(count = n()) %>%
+    ggplot(aes(x = tmin, y = count)) +
+    geom_point( colour = 'purple') +
+    facet_wrap(~ is_substantial) +
+    geom_smooth(method = "lm") +
+    labs(x = "Minimum temperature", y = "Trip count", title = "Trip count by minimum temperature for each day")
```
hannahcha417, line 164: Great job with these plots!!

week1/plot_trips.R
```diff
@@ -47,16 +112,89 @@ trips_with_weather <- inner_join(trips, weather, by="ymd")
...
+trips_with_weather %>%
+    mutate(hour = hour(starttime)) %>%
+    group_by(date, hour) %>%
+    summarize(count=n()) %>%
+    group_by(hour) %>%
+    summarize(avg = mean(count), sd = sd(count)) %>%
+    ggplot(aes(x = hour, y = avg)) +
+    geom_pointrange(aes(ymin = avg - sd, ymax = avg + sd)) +
+    coord_flip() +
+    labs(x = "Hour of Day", y = "Average Trips", title = "Average Trips by Hour of Day")
```
hannahcha417, line 188: I would actually not flip the coordinates here; it makes sense for the x axis to be the hour and the y axis to be the number of trips!

week1/plot_trips.R
```diff
@@ -47,16 +112,89 @@ trips_with_weather <- inner_join(trips, weather, by="ymd")
...
+trips %>%
+    mutate(date = as.Date(starttime), day_of_week = wday(date)) %>%
+    group_by(date, day_of_week) %>%    
+    summarize(count=n()) %>%
+    group_by(day_of_week) %>%
+    summarize(avg = mean(count), sd = sd(count)) %>%
+    ggplot(aes(x = day_of_week, y = avg)) +
+    geom_pointrange(aes(ymin = avg - sd, ymax = avg + sd)) +
+    coord_flip() +
+    labs(x = "Day Of Week", y = "Average Trips", title = "Average Trips by Day of Week")
```
hannahcha417, line 200: I think in this case, it would probably be best to simply create the graph you did in the previous problem, but use facet_wrap to show separate graphs for each day of the week! Currently, the information you're showing is a bit different; you're getting the average trips per day of the week, rather than the average trips per hour for each given day of the week, if that makes sense!

hannahcha417:
Overall, great job on this assignment and great graphs! I left a few comments on places/suggestions in places where your graphs could be improved. Great job!

PR #458, opened 2026-06-01
Day4 Assignment

Ebeene01:
Day4 Assignment Completed


week1/plot_trips.R
```diff
@@ -18,25 +18,78 @@ load('trips.RData')
...
 # plot the distribution of trip times across all rides (compare a histogram vs. a density plot)
+ggplot(trips, aes(x = tripduration)) +
+    scale_x_log10(label = comma) +
+    scale_y_continuous(label = comma) +
+    geom_histogram()
+
+ggplot(trips, aes(x = tripduration)) +
+    scale_x_log10(label = comma) +
+    scale_y_continuous(label = comma) +
+    geom_density(fill="pink")
```
hannahcha417, line 30: Here, I would recommend filtering out unusually large values (e.g. 1 million seconds). Then, you wouldn't need to use a log scale to represent the x axis!

week1/plot_trips.R
```diff
@@ -18,25 +18,78 @@ load('trips.RData')
...
 
 # plot the distribution of trip times by rider type indicated using color and fill (compare a histogram vs. a density plot)
+ggplot(trips, aes(x = tripduration, color=usertype, fill=usertype)) +
+    scale_x_log10(label = comma) +
+    geom_histogram()
+
+ggplot(trips, aes(x = tripduration, color=usertype, fill=usertype)) +
+    scale_x_log10(label = comma) +
+    geom_density()
```
hannahcha417, line 38: For both of these, it's a bit hard to see the different types on top of each other as histogram/density plots; I would recommend separating these out into two separate graphs using facet_wrap!

week1/plot_trips.R
```diff
@@ -18,25 +18,78 @@ load('trips.RData')
...
+total_trips <- trips %>%
+    mutate(age = lubridate::year(Sys.Date()) - birth_year)%>%
+    group_by(age,gender)%>%
+    summarize(num_trips = n())
+
+ggplot(total_trips, aes(x = age, y = num_trips, color=gender)) +
+    scale_y_continuous(label = comma) +
+    xlab('Age')+
+    ylab('Number of trips')+
+    geom_line()
```
hannahcha417, line 61: Nice graph! Something that can make your data easier to see is setting a limit on the y axis!

week1/plot_trips.R
```diff
@@ -18,25 +18,78 @@ load('trips.RData')
...
 
 ########################################
 # plot weather data
 ########################################
 # plot the minimum temperature (on the y axis) over each day (on the x axis)
+ggplot(weather, aes(x = date, y = tmin)) +
+    xlab('Date')+
+    ylab('Minimum temperature')+
+    geom_point()
```
hannahcha417, line 84: I would recommend using ymd here for a cleaner x axis!

week1/plot_trips.R
```diff
@@ -47,16 +100,59 @@ trips_with_weather <- inner_join(trips, weather, by="ymd")
...
 
 # add a smoothed fit on top of the previous plot, using geom_smooth
+trips_with_weather%>%
+    mutate(precip=ifelse(prcp>mean(prcp), "T", "F"))%>%
+    group_by(ymd,precip)%>%
+    summarise(count=n(), mean=mean(tmin))%>%
+    ggplot(aes(x=mean, y=count))+
+    geom_point()+
+    geom_smooth(method="lm")
```
hannahcha417, line 125: In both the graphs above, you added the column correctly, but you are not actually using that information! A simple way to do this is to set `color = precip` in your ggplot params so each label gets represented as a different color

week1/plot_trips.R
```diff
@@ -47,16 +100,59 @@ trips_with_weather <- inner_join(trips, weather, by="ymd")
...
 # hint: use the wday() function from the lubridate package
+trips_with_weather%>%
+    mutate(weekday=wday(ymd, label = TRUE), hour= hour(starttime), day=floor_date(ymd,unit = "day"))%>%
+    group_by(day,hour,weekday)%>%
+    summarise(count=n())%>%
+    group_by(hour, weekday)%>%
+    summarise(mean=mean(count), stds=sd(count))%>%
+    ggplot(aes(x=hour))+
+    geom_ribbon(aes(ymin= mean-stds, ymax=mean+stds, alpha=0.5))+
+    facet_wrap(~ weekday, scale="free")
```
hannahcha417, line 158: Nice graphs! For both of the above plots, don't forget to include geom_line() in addition to geom_ribbon; otherwise, you're only plotting the standard deviation!

week1/plot_trips.R
```diff
@@ -18,25 +18,78 @@ load('trips.RData')
...
+    geom_point()
 
 # plot the minimum temperature and maximum temperature (on the y axis, with different colors) over each day (on the x axis)
 # hint: try using the pivot_longer() function for this to reshape things before plotting
 # (you can skip this and come back to it tomorrow if we haven't covered reshaping data yet)
+pivot_longer(weather, names_to = "min_max", values_to = "Temperature", 10:11) %>%
+ggplot(aes(x = date, y = Temperature, group=min_max, color=min_max)) +
+    xlab('Date')+
+    ylab('Daily Temperature Level ')+
+    geom_line()
```
hannahcha417, line 92: Right now this is giving an out of index error; weather only has 7 columns, but you're trying to access 10 and 11! I would recommend using the actual column names in this argument like so: cols = c(tmin, tmax)

hannahcha417:
Overall, good job on the assignment and nice graphs! I left a few comments on areas for improvement and smaller comments for suggestions; let me know if you have any questions!

PR #461, opened 2026-06-02
Completed day 4's assignment 

PrestBlake25:
I just completed day 4's assignment.


week1/plot_trips.R
```diff
@@ -18,26 +18,53 @@ load('trips.RData')
 ########################################
 
 # plot the distribution of trip times across all rides (compare a histogram vs. a density plot)
-
+ggplot(trips, aes(x = tripduration)) + scale_x_log10(label = comma) + geom_histogram(bins = 200)
+ggplot(trips, aes(x = tripduration)) + scale_x_log10(label = comma) + geom_density()
 # plot the distribution of trip times by rider type indicated using color and fill (compare a histogram vs. a density plot)
```
hannahcha417, line 23: For these graphs, I would recommend filtering out some of the unusually large values (e.g. 1 million seconds)! And that way, you won't have to use a log scale x axis

week1/plot_trips.R
```diff
@@ -18,26 +18,53 @@ load('trips.RData')
...
-
+ggplot(trips, aes(x = tripduration, fill = usertype, color = usertype)) + scale_x_log10(label = comma)+ geom_histogram(bins = 200)
+ggplot(trips, aes(x = tripduration, color = usertype)) + scale_x_log10(label = comma)+ geom_density()
 # plot the total number of trips on each day in the dataset
```
hannahcha417, line 26: Histogram/density plots of two different types are quite hard to see on the same graph! I would recommend splitting them out into different graphs with facet_wrap, or maybe even reducing the opacity of one or the other to make it easier to see!

week1/plot_trips.R
```diff
@@ -18,26 +18,53 @@ load('trips.RData')
...
+new_trips %>% group_by(start_date) %>% summarize(num_trips = n()) %>% ggplot(aes(x=start_date, y = num_trips)) + geom_line()
 
 # plot the total number of trips (on the y axis) by age (on the x axis) and gender (indicated with color)
 
+newer_trips <- new_trips %>% mutate (
+                age = 2014 - birth_year
+            )
+newer_trips %>% group_by(age, gender) %>% summarize(num_trips = n()) %>% ggplot(aes(x = age, y = num_trips, color = gender)) + geom_line()
+
 # plot the ratio of male to female trips (on the y axis) by age (on the x axis)
```
hannahcha417, line 42: You could lower the y axis here to make it easier to see!

week1/plot_trips.R
```diff
@@ -18,26 +18,53 @@ load('trips.RData')
...
+                                            mutate(m_t_f_ratio = Male / Female) %>%
+                                            ggplot(aes(x = age, y = m_t_f_ratio)) + scalegeom_point()
+
 ########################################
 # plot weather data
 ########################################
 # plot the minimum temperature (on the y axis) over each day (on the x axis)
 
+ggplot(weather, aes(x = date, y = tmin)) + geom_point()
+
```
hannahcha417, line 59: You can use x = ymd for a cleaner x axis!

week1/plot_trips.R
```diff
@@ -48,15 +75,60 @@ trips_with_weather <- inner_join(trips, weather, by="ymd")
...
 # you'll need to decide what constitutes "substantial precipitation" and create a new T/F column to indicate this
 
+trips_with_weather %>% group_by(ymd, tmin) %>% summarise(num_trips = n(), avg_precp = mean(prcp)) %>%
+                        ggplot(aes(x = tmin, y = num_trips, color = avg_precp)) + geom_point() +
+                    labs(
+                        title = "Min Temperature vs Number Of Trips",
+                        x = "Min Temperature",
+                        y = "Number of Trips"
+                    ) + geom_smooth()
+
```
hannahcha417, line 98: Right now, you're adding a color to results based on the mean precipitation, but we were thinking you could split up results using a precipitation threshold that you can specify! This would involve mutating to create a new binary column, and then using that column to either specify the color of the point or facet_wrap based on it!

week1/plot_trips.R
```diff
@@ -48,15 +75,60 @@ trips_with_weather <- inner_join(trips, weather, by="ymd")
...
+ggplot(weekday_avg_std, aes(x = weekday, y = avg_num_trips)) +
+geom_line() +
+geom_ribbon(aes(ymin = avg_num_trips - standard_devi,
+                ymax = avg_num_trips + standard_devi),
+                alpha = 0.3) +
+        labs (title = "Weekday vs Number Of Trips (With Standard Deviation)",
+                        x = "Weekday",
+                        y = "Average Number of Trips"
+            )
```
hannahcha417, line 134: For this graph, I think we intended for you to use facet_wrap to split up the average trips per hour across the day, for each day of the week, creating 7 different graphs for each day. Currently, you're plotting the average number of trips per day of the week, which is subtly different!

hannahcha417:
Great job overall and great graphs! I included a few comments/suggestions here and there, but overall, great work! Let me know if you have any questions on some of the comments I left

PR #463, opened 2026-06-02
Visualization Asssignment Day 4 

faridabalogun:
So sorry it too so long, i wanted to make the code easier to read


week1/plot_trips.R
```diff
@@ -18,25 +18,60 @@ load('trips.RData')
 ########################################
 
 # plot the distribution of trip times across all rides (compare a histogram vs. a density plot)
+trips %>%
+ggplot(aes(x=tripduration)) + geom_histogram()+
+scale_x_log10(label = comma )+ xlab('Trip Duration') +ylab('Count')
+
+ggplot(trips,aes(x=tripduration))+ geom_density(fill = "grey"  )+
+scale_x_log10(label = comma )+ xlab('Trip Duration') +ylab('Count')
```
hannahcha417, line 27: For these two, I would recommend filtering out unusually large values (e.g. 1 million seconds)! That way, you don't have to use a log scale for the x axis

week1/plot_trips.R
```diff
@@ -18,25 +18,60 @@ load('trips.RData')
...
 
 # plot the distribution of trip times by rider type indicated using color and fill (compare a histogram vs. a density plot)
+ggplot(trips, aes(x = tripduration  , color = usertype, fill =  usertype ))  + geom_histogram() +scale_x_log10(label = comma )+
+xlab('Trip Duration') +ylab('Count of Trips')
+ggplot(trips, aes(x = tripduration  , color = usertype, fill =  usertype ))  + geom_density() +scale_x_log10(label = comma )+
+xlab('Trip Duration') +ylab('Count of Trips')
```
hannahcha417, line 33: For both of the above, I would recommend separating out the user types using facet wrap! For histograms and density plots, it's a bit difficult to see the two different types when they're on top of each other!

week1/plot_trips.R
```diff
@@ -18,25 +18,60 @@ load('trips.RData')
...
+ggplot(aes(x = day , y = tmin))+ geom_point()+scale_y_log10(label = comma )+ xlab('Day') +ylab('Minimum Temperature')
 
 # plot the minimum temperature and maximum temperature (on the y axis, with different colors) over each day (on the x axis)
 # hint: try using the pivot_longer() function for this to reshape things before plotting
 # (you can skip this and come back to it tomorrow if we haven't covered reshaping data yet)
+weather %>%
+select(date,tmin,tmax) %>%
+pivot_longer(names_to ="temp", values_to = "temp_values", 2:3 )%>%
+ggplot(aes(x = date, y = date ))+geom_point() +  xlab('Day') +ylab('Minimum And Maximum Temperature')
+
```
hannahcha417, line 74: In this case, you should plot it like y = temp_values, color = temp, so that you can see both min and max cleanly on the same grpah. Also if you set x = ymd, you'll get a cleaner x axis!

week1/plot_trips.R
```diff
@@ -47,16 +82,47 @@ trips_with_weather <- inner_join(trips, weather, by="ymd")
...
 # repeat this, but now split the results by day of the week (Monday, Tuesday, ...) or weekday vs. weekend days
 # hint: use the wday() function from the lubridate package
+trips_with_weather %>% 
+mutate(weekday = wday(starttime,label = TRUE),hour = hour(starttime),day = floor_date(ymd,unit="day"))%>% 
+group_by(day,hour,weekday)%>%summarize (count= n())%>%
+group_by(hour,weekday)%>%
+summarize (mean = mean(count),sd = sd(count))%>%
+ggplot(aes( x= hour))+ 
+geom_ribbon(aes(ymin = mean - sd , ymax  = sd + mean,alpha = 0.25))+facet_wrap(~weekday,scale ="free")
++ xlab(' Hour') +ylab('SD and Mean of Trips')
```
hannahcha417, line 128: Great graphs! Just an FYI that you'll also want to include geom_line() to include the actual data, not just its standard deviation!

hannahcha417:
Overall, great job! I left a few comments/suggestions in places; let me know if you have any questions about these! Your graphs are great!

PR #475, opened 2026-06-03
assignments up to week 2 day 2

pennywa:


week2/sampling-distributions.R
```diff
@@ -80,19 +109,45 @@ library(tidyverse)
...
+ x.avgs <- rep(0,10^4)
+  for(i in 1:10^4)
+{
+x.group <- sample(pop2$bmi,150)
+x.avgs[i] <- mean(x.group)
+}
+mean(x.avgs)
+# 24.98105
+# basically it's 25. values are slightly different when i run it multiple times
+
```
hannahcha417, line 131: You could also use the replicate function to do this like this: `sample_means <- replicate(10000, mean(sample(pop2$bmi, size = 150)))`

hannahcha417:
Nice work!!

PR #478, opened 2026-06-03
Day 7 Assignment

addinar:
Completed week2/hypothesis_testing.R


week2/hypothesis-testing.R
```diff
@@ -7,19 +7,73 @@ magnets <- read_csv("http://pluto.huji.ac.il/~msby/StatThink/Datasets/magnets.cs
...
+magnets %>%
+    ggplot(aes(x=active, y=change)) +
+    geom_boxplot() + 
+    labs(
+        x="Patient Categories",
+        y="Change in Response"
+    )
+# The boxplot for the change in response for patients who received the active magnets has 0 outliers
+# The boxplot for the change in response for patients who received the placebo has 3 outliers
+
```
hannahcha417, line 76: Nice job figuring out the boxplot!

week2/hypothesis-testing.R
```diff
@@ -33,12 +87,99 @@ magnets <- read_csv("http://pluto.huji.ac.il/~msby/StatThink/Datasets/magnets.cs
 #    of size n = 100 from the Normal(3, 2) distribution. Compute the expectation
 #    and the variance of the sample average and of the sample median. Which of
 #    the two estimators has a smaller mean square error?
+mu <- 3
+sd <- 2
+n <- 100
```
hannahcha417, line 92: Close! The second parameter to Normal() is actually the variance, not the standard deviation. So, you're going to have to take the square root of it to get the standard deviation! Your answers below are slightly off because of this, but super easy to fix since all you need to do is change the sd value to sqrt(2)!

week2/hypothesis-testing.R
```diff
@@ -33,12 +87,99 @@ magnets <- read_csv("http://pluto.huji.ac.il/~msby/StatThink/Datasets/magnets.cs
...
+variance_median_normal
+# The variance when using median as an estimator is 0.06261095
+
+# Mean squared error of using median as an estimator:
+mse_median_normal <- (expectation_median_normal - mu)^2 + variance_median_normal
+mse_median_normal
+# The MSE is 0.06261095
+
+# Therefore, using mean as an estimator for the normal distribution has a slightly smaller MSE, making it only slightly more accurate
+
```
hannahcha417, line 132: That's right!

week2/hypothesis-testing.R
```diff
@@ -146,11 +397,65 @@ ex2 <- read_csv("http://pluto.huji.ac.il/~msby/StatThink/Datasets/ex2.csv")
...
+    delta <- proprotion_yawn_treatment - proportion_yawned_control
+    delta
+    # The difference is 0.04411765
+
 # (c) Estimate the p-value using the figure and determine the conclusion of
 #     the hypothesis test.
 
+    # From the null distribution, the p-value of 0.04411765 would be around ~0.25
+    # This is the probability of getting a difference this extreme if the null hypothesis were true.
+
```
hannahcha417, line 458: Exactly, so we failed to reject the null hypothesis!

week2/hypothesis-testing.R
```diff
@@ -175,5 +480,31 @@ ex2 <- read_csv("http://pluto.huji.ac.il/~msby/StatThink/Datasets/ex2.csv")
...
+    mean_active_sample <- run_average(29, mu, s1)
+    mean_placebo_sample <- run_average(21, mu, s2)
+    test_stat <- (mean_active_sample - mean_placebo_sample) / sqrt((s1^2)/29 + (s2^2)/21)
+}
+
+test_stat_sample <- replicate(1e4, run_samples())
+
+quantile(test_stat_sample, c(0.025, 0.975))
+# 2.5%     97.5% 
+# -1.967156  1.937554
```
hannahcha417, line 501: This is not quite right; the denominator should use the sample variances computed inside run_samples(). You should use rnorm to get the mean and variances (X1 = rnorm (29, mu, s1), for example), and then get the mean and the variance of X1 to use in the formula, if that makes sense!

hannahcha417:
Great work on this assignment! I left a few comments on small errors that can easily be fixed, but otherwise, good job!

PR #484, opened 2026-06-04
Week 2, day 3 and 4

ChanaRapp:
Thank you!


week2/hypothesis-testing.R
```diff
@@ -33,11 +41,18 @@ magnets <- read_csv("http://pluto.huji.ac.il/~msby/StatThink/Datasets/magnets.cs
 #    of size n = 100 from the Normal(3, 2) distribution. Compute the expectation
 #    and the variance of the sample average and of the sample median. Which of
 #    the two estimators has a smaller mean square error?
-#
+mean(replicate(100000, mean(rnorm(n=100, mean=3, sd=2))))     # 2.999544
+var(replicate(100000, mean(rnorm(n=100, mean=3, sd=2))))      # 0.04030229
+mean(replicate(100000, median(rnorm(n=100, mean=3, sd=2))))   # 2.999957 
+var(replicate(100000, median(rnorm(n=100, mean=3, sd=2))))    # 0.06175777
 # 2. Simulate the sampling distribution of average and the median of a sample
```
hannahcha417, line 48: Close! You're actually going to want to set sd = sqrt(2), because the second parameter passed to the Normal() function is the variance, not the sd! So you will have to convert it before

week2/hypothesis-testing.R
```diff
@@ -175,5 +204,14 @@ ex2 <- read_csv("http://pluto.huji.ac.il/~msby/StatThink/Datasets/ex2.csv")
 #    measurements is Normal and there are 29 patients in the first group and 21
 #    in the second. Find the interval that contains 95% of the sampling
 #    distribution of the statistic.
+sampling_dist <- replicate(100000, (mean(rnorm(29, mean = 3.5, sd = 3)) - mean(rnorm(21, mean = 3.5, sd = 1.5))) / 
+                sqrt(((sd(rnorm(29, mean = 3.5, sd = 3))**2)/29) + (sd(rnorm(21, mean = 3.5, sd = 1.5))/21)))
+quantile(sampling_dist, .025)
+quantile(sampling_dist, .975)
+# Interval [-2.116607, 2.127696]
 # 2. Does the observed value of T (computed from the "magnets" data) fall
```
hannahcha417, line 212: One thing you're doing here is computing the numerator and denominator from different data, because you're calling the rnorm another time. You're going to want to save the samples and use the same ones in both the numerator and the denominator, and calculate the mean and variance from the same data

hannahcha417:
Great work on these assignments! I left a few comments/suggestions in some places :)

ChanaRapp:
Thank you for clarifying! I fixed those problems.

PR #485, opened 2026-06-05
Day 8 Assignment

addinar:
Completed week2/power-effect-replication.R


week2/power-effect-replication.R
```diff
@@ -26,10 +26,75 @@ magnets <- read_csv("http://pluto.huji.ac.il/~msby/StatThink/Datasets/magnets.cs
...
+# 95 percent confidence interval:
+#  0.3768845 1.8135916
+# sample estimates:
+# mean of x 
+#  1.095238 
+
+# Conclusions:
+# again, reject the null hypothesis since the p-value 0.004702 < 0.05
+# There is a real placebo effect
+
```
hannahcha417, line 97: Wow, good work doing the step by step, manual way to get the p-value!!

week2/regression.R
```diff
@@ -8,16 +8,73 @@
...
+height <- function(shoulder_girth) {
+    m * shoulder_girth + b
+}
+
 # b. Intepret the slope and the intercept in this context.
+
+# The slope 0.6079749 is the proportion between height and shoulder girth in the least squares line.
+
+# The intercept 105.6171 what the height would approximately be if the shoulder girth was 0.
+
```
hannahcha417, line 41: More explicitly, it would mean that for every 1 cm increase in shoulder girth, the height is predicted to increase by 0.608 cm on average

hannahcha417:
Great job on both of these assignments!! :)

PR #486, opened 2026-06-05
Week2/Days 3, 4 & 5

Ebeene01:
Assignment Completed!!


week2/hypothesis-testing.R
```diff
@@ -146,10 +261,34 @@ ex2 <- read_csv("http://pluto.huji.ac.il/~msby/StatThink/Datasets/ex2.csv")
...
 #     the hypothesis test.
+ # The observed value 0.044 is very close to 0 and lies near the center of the distribution
+ # Based on the alternative hypothesis, the p-value is the proportion with 
+ # ptrtmt - pˆctrl >= 0.044 
+ # Based on the histogram, almost half of tge values are >0, and since 
+ # 0.044 is only slightly above 0, a large fraction of simulations are at least this large
+ # We can then estimate that p-value is around 0.4
+ # p-value > 0.05 --> can't reject the null hypothesis
+ # We are not convinced that having someone yawn nearby increases the likelihood of yawning
+
```
hannahcha417, line 291: The observed difference is actually not the p value! The p value can be estimated from the figure, and it comes out to around 0.54 (which is not a significant difference)

hannahcha417:
Great job on these assignments!! I left a small comment, but your work looks great otherwise :)

PR #492, opened 2026-06-06
Assignment for 3rd and 4 th day #week2

n-oli:
Hi Hannah, I am submitting both my 3rd and 4th day assignment through this pool request.


week2/hypothesis-testing.R
```diff
@@ -7,20 +7,38 @@ magnets <- read_csv("http://pluto.huji.ac.il/~msby/StatThink/Datasets/magnets.cs
...
+
+mean(magnets$change)#3.5
+
 # 2. Is the variable "active" a factor or a numeric variable?
+
+summary(magnets)
+str(magnets)
+# No it is not a factor, it is a character variable 
+
 # 3. Compute the average value of the variable "change" for the patients that
```
hannahcha417, line 19: It actually is a factor variable, although it is represented by numbers!

week2/hypothesis-testing.R
```diff
@@ -33,12 +51,46 @@ magnets <- read_csv("http://pluto.huji.ac.il/~msby/StatThink/Datasets/magnets.cs
...
+    mean_stat[n] <- mean(X)
+    median_stat[n] <- median(X)
+}
+mean(mean_stat)
+mean(median_stat)
+var(mean_stat)
+var(median_stat)
+#since var(median_stat) > var(mean_stat) so mean has smaller mean squared error
+
 # 2. Simulate the sampling distribution of average and the median of a sample
```
hannahcha417, line 72: In this case, you actually want to set std = sqrt(2), not just 2. The second parameter to the Normal() function is actually the variance, so you'll have to take the square root to convert!

hannahcha417:
Nice work on these assignments!!

PR #494, opened 2026-06-08
Assignments up to Week 2 Day 5 Part 1

pennywa:
Most recently completed is regression.R from Week 2 Day 5 Part 1.

Need to continue from Week 2 Day 5 Part 2.


hannahcha417:
Great work on these assignments!!

PR #500, opened 2026-06-08
week 3 1st assignment

AlexandruSoroiu:
submitted regression-pt2.R


week2/regression-pt2.R
hannahcha417, line 9: For this problem, weight and height are columns 23 and 24, respectively!

hannahcha417:
Great job on this assignment!

PR #502, opened 2026-06-08
Week3/Day1

Ebeene01:
Today's Assignment Completed!!


week2/regression-pt2.R
hannahcha417, line 9: The weight and height columns are 23 and 24, respectively!

hannahcha417:
Nice work on this assignment!!

PR #511, opened 2026-06-10
Finished coffee_cancer.R

pennywa:
created new file
did parts a and b in that file


week3/coffee_cancer.R
```diff
@@ -0,0 +1,26 @@
...
+b) Comment on challenges you might face in running this experiment."
+
+# Ans a)
+power.prop.test(p1 = 0.40, p2 = 0.40001, sig.level = 0.10, power = 0.80, alternative = "one.sided") 
+
+# Ans b)
+# There might be a problem with the sample size and the global population that boils down to whether we have enough people or not.
+# How long is the study going to run for? Will participants drop out of the study simply because it's too long?
+# Other noise that people will be exposed to like environment factors, lifestyle changes other potential carcinogens like air polllution, smoking, etc.
+# How do we check that the people are sticking 100% (never drinking coffee group; what if one person cheats and takes 1 sip?)
```
hannahcha417, line 26: These last few things are all good things to worry about! I think the main problem with the experiment is the first one you pointed out, where the sample size would make an experiment like this infeasible. Your other worries are also good answers, and I believe medical studies have picked up on ways to account for these things, since they encounter these in their regular experiments (although I'm not an expert in this by any means)

hannahcha417:
Great work on this assignment!

PR #512, opened 2026-06-10
Fix download script

pennywa:
Changed http to https on line 8 (the URL) also ml-10m (make sure ml is lowercase)

url=https://files.grouplens.org/datasets/movielens/ml-10m.zip
Added -L flag to the curl command on line 11 (to follow redirects)
curl -L -o movielens_10M.zip $url
-L before the -o (order matters)

If ratings.csv and movies.tsv files are empty
Run rm ratings.csv then cat ratings.dat | sed 's/::/,/g' > ratings.csv
Run rm movies.tsv then cat movies.dat | sed 's/::/\t/g' > movies.tsv
^Now the files have content


jhofman:
thanks @pennywa, this works! i can't merge the PR itself because it has *all* your code changes from the summer, but i cherry picked commit 0d4ef9 and applied it to the main branch so that everyone should be able to get the working script now.

PR #533, opened 2026-06-15
Completed predict_citibike and ngrams

ChanaRapp:
Please see my completed work from last week and today's assignment in week4. Thank you!


hannahcha417:
Nice work on the ngrams assignment!! Will come back to this PR to grade the citibike assignment

week3/predict_citibike.Rmd
```diff
@@ -40,6 +71,47 @@ Add features that might help: day of week, month, precipitation, etc.
...
+trips_per_day_train <- 
+    trips_per_day_train %>%
+    mutate(month = mday(ymd))
+
+trips_per_day_validate <- 
+    trips_per_day_validate %>%
+    mutate(month = mday(ymd))
+
+model8 <- lm(num_trips ~ poly(tmin, degree = 1) + snow + prcp + weekend + tmax + month, data=trips_per_day_train)
+validate_err_8 <- sqrt(mean((predict(model8, trips_per_day_validate) - trips_per_day_validate$num_trips)^2))
```
hannahcha417, line 115: Great work testing a lot of different models!

week4/test_citibike_predictions.Rmd
```diff
@@ -0,0 +1,53 @@
...
+    xlab("Date") +
+    ylab("Predicted and Actual Number of Trips")
+```
+
+## Results
+
+My RMSE in my cross validation was 4608.399, while it was 8578.844 on the 2015 data.
+I think that my model could have fit better than it did, but considering the risks of 
+overfitting because people's behavior can never be fully predicted, I think this model 
+did pretty well, with an rsquared value of 0.6841944.
```
hannahcha417, line 53: Agreed! And again, as Jake mentioned, it might very well be the case that last year's data can only predict so much about the current year.

hannahcha417:
Nice work on the citibike assignment! You did great at testing a number of different models and validating the results :)

